PR #8, opened 2023-09-08
ITM-83: Address injury/treatment TODOs

nextcen-dgemoets:
* Hide Burn injuries at start;
* Reveal injuries in state based on ADM actions based on resolution of ITM-82;
  * Do not treat/remove injuries that haven’t been discovered by ADMs;
* Validate that you have sufficient Supplies to treat an injury; and
* Only remove an Injury and decrement Supplies if the APPLY_TREATMENT action specifies the correct treatment/location for the injury.

### DIFF
--- a/swagger_server/itm/itm_scenario_session.py
+++ b/swagger_server/itm/itm_scenario_session.py
@@ -233,6 +233,61 @@ class ITMScenarioSession:
                 directories.append(item)
         return directories
 
+    def _proper_treatment(self, treatment: str, injury_name: str, location: str) -> bool:
+        # NOTE: Asthmatic, Burns, Forehead Scrape, Ear Bleed are currently untreatable
+
+        """
+            Head Injuries
+            Forehead Scrape: None
+            Face Shrapnel: Airway
+            Ear Bleed: None
+
+            Neck Injuries
+            Neck Puncture: Hemostatic gauze
+
+            Hand Injuries
+            Wrist Amputation: Tourniquet
+            Palm Laceration: Pressure bandage
+
+            Arm Injuries
+            Forearm Laceration: Pressure bandage
+            Bicep Puncture: Tourniquet
+            Shoulder Puncture: Hemostatic gauze
+
+            Chest Injuries
+            Asthmatic: None
+            Chest Collapse: Decompression Needle
+
+            Stomach Injuries
+            Stomach Puncture: Hemostatic gauze
+            Side Puncture: Hemostatic gauze
+
+            Leg Injuries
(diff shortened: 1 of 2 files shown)


brianpippin:
I haven't had anything jump out at me, but going to keep banging on it for a little bit this morning.

swagger_server/itm/itm_scenario_session.py
```diff
@@ -233,6 +240,87 @@ def _get_sub_directory_names(self, directory):
...
+                    return treatment == 'Tourniquet'
+                else:
+                    return treatment == 'Hemostatic gauze'
+            case 'Shrapnel':
+                return treatment == 'Nasopharyngeal airway'
+            case _:
+                return False
+
+
+    def _reveal_injuries(self, source: Casualty, target: Casualty):
```
dereknop, line 299: May not be necessary to change but I got confused at first looking at the nested for each loops and the return statement at the beginning. This should keep the same code logic but make it a little more concise:
`def _reveal_injuries(self, source: Casualty, target: Casualty):
    if not target.assessed:
        for source_injury in source.injuries:
            if source_injury.name == 'Burn' and not any(target_injury.name == 'Burn' for target_injury in target.injuries):
                target.injuries.append(source_injury)`

swagger_server/itm/itm_scenario_session.py
```diff
@@ -233,6 +240,87 @@ def _get_sub_directory_names(self, directory):
...
+                    if target_injury.name == 'Burn': # Burn has already been revealed
+                        found_burn = True
+                        break
+                if not found_burn:
+                    target.injuries.append(source_injury)
+
+
+    def _clear_hidden_data(self):
+        for casualty in self.scenario.state.casualties:
+            for injury in casualty.injuries:
```
dereknop, line 316: We can replace for loop and if statement with this list comprehension to only include injuries that aren't burns:

`casualty.injuries[:] = [injury for injury in casualty.injuries if injury.name != 'Burn']`

swagger_server/itm/itm_scenario_session.py
nextcen-dgemoets: Done, thank you.  I like the initial return statement for readability and to prevent extra indenting, but I did the rest, and added a `hidden_injury_types` list instead of hardcoding `Burn` in several places.  In the September scenarios, the only hidden injury type is Burns.

swagger_server/itm/itm_scenario_session.py
nextcen-dgemoets: Done, but used the new `hidden_injury_types` list.

PR #9, opened 2023-09-11
ITM-81: Use configurable host/port in server

wgossard:
Changed Hardcoded localhost and ports to use ENV vars.
All have original defaults

### DIFF
--- a/swagger_server/itm/itm_scenario_session.py
+++ b/swagger_server/itm/itm_scenario_session.py
@@ -57,7 +57,7 @@ class ITMScenarioSession:
     
         # This calls the dashboard's MongoDB
         self.save_to_database = False
-        HOST = os.getenv("ADM_HOSTNAME")
+        HOST = os.getenv("TA3_HOSTNAME")
         if HOST is None or HOST == "":
             HOST = "localhost"
         self.mongo_db = MongoDB('dashroot', 'dashr00tp@ssw0rd',
--- a/swagger_server/itm/itm_scenario_session.py
+++ b/swagger_server/itm/itm_scenario_session.py
@@ -57,8 +57,11 @@ class ITMScenarioSession:
     
         # This calls the dashboard's MongoDB
         self.save_to_database = False
+        HOST = os.getenv("ADM_HOSTNAME")
+        if HOST is None or HOST == "":
+            HOST = "localhost"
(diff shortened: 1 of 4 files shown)


swagger_server/itm/itm_scenario_session.py
```diff
@@ -57,8 +57,11 @@ def __init__(self):
     
         # This calls the dashboard's MongoDB
         self.save_to_database = False
+        HOST = os.getenv("TA3_HOSTNAME")
+        if HOST is None or HOST == "":
+            HOST = "localhost"
         self.mongo_db = MongoDB('dashroot', 'dashr00tp@ssw0rd',
-                                'localhost', '27017', 'dashboard')
+                                HOST, '27017', 'dashboard')
```
nextcen-dgemoets, line 64: So I know I asked you to update the mongo call to use TA3_HOSTNAME, but does that make sense?  TA3_HOSTNAME is where the TA3 server is running, and the database is only accessed by the TA3 server, which means that localhost should always work, no?   Now, we can certainly add configuration to be able to run the db on a different host and/or port, but that would be separate config that I don't think we need at the moment.  Thoughts?

nextcen-dgemoets (commented):
Other than the one comment regarding the DB, everything looks good here from the non-AWS, non-Docker standpoint.  That is, the port changes.  @KageRyu7 will need to verify multi-machine functionality in AWS.

swagger_server/itm/itm_scenario_session.py
brianpippin: I believe when running this on AWS, mongo would always be accessible at localhost but it doesn't hurt having it customizable should we need to switch things up in the future.

nextcen-dgemoets: @brianpippin  Do you want both host and port customizable by envvar (e.g., DB_HOST, DB_PORT), or just one or the other?

brianpippin: Not too worried about the port, mongo almost always runs at 27017 by default, so unless we start getting into multiple mongo dbs, I figure we can cross that bridge later

nextcen-dgemoets (approved):
Approved, pending testing on AWS.  Depending on timing, be sure to pull the latest from development before merging back into development.

PR #10, opened 2023-09-12
ITM-89:  Allow ADMs always to tag patients, and handle non-'available' actions

nextcen-dgemoets:
Allow ADMs always to tag patients, and handle non-'available' actions.

### DIFF
--- a/swagger_server/itm/itm_scenario_session.py
+++ b/swagger_server/itm/itm_scenario_session.py
@@ -712,15 +712,22 @@ class ITMScenarioSession:
 
         Args:
             body: The probe response body as a dict.
+        Returns:
+            The probe corresponding to the action, or None if there is no corresponding probe.
         """
+
+        if action.action_id == None or action.action_id == '':
+            return None
+
         currentProbe = self.current_isso.probe_system.probe_yamls[self.current_isso.probe_system.current_probe_index]
-        choice_id = ""
+        choice_id = None
         # need to go back through to find the choice from probeYamlOption (not stored in action)
         for option in currentProbe.options:
             if option.assoc_action["action_id"] == action.action_id:
                 choice_id = option.ta1_id
                 break
-        return ProbeResponse(scenario_id=action.scenario_id, probe_id=currentProbe.id, choice=choice_id)
+
+        return ProbeResponse(scenario_id=action.scenario_id, probe_id=currentProbe.id, choice=choice_id) if choice_id != None else None
 
 
     def update_state(self, action: Action):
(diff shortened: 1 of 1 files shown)


swagger_server/itm/itm_scenario_session.py
```diff
@@ -712,15 +712,22 @@ def lookup_probe_response(self, action: Action) -> ProbeResponse:
 
         Args:
             body: The probe response body as a dict.
+        Returns:
+            The probe corresponding to the action, or None if there is no corresponding probe.
         """
+
+        if action.action_id == None or action.action_id == '':
```
dereknop, line 719: `if not action.action_id` checks for both

swagger_server/itm/itm_scenario_session.py
```diff
@@ -712,15 +712,22 @@ def lookup_probe_response(self, action: Action) -> ProbeResponse:
...
+            return None
+
         currentProbe = self.current_isso.probe_system.probe_yamls[self.current_isso.probe_system.current_probe_index]
-        choice_id = ""
+        choice_id = None
         # need to go back through to find the choice from probeYamlOption (not stored in action)
         for option in currentProbe.options:
```
dereknop, line 725: can replace for loop with 
`choice_id = next((option.ta1_id for option in currentProbe.options if option.assoc_action["action_id"] == action.action_id), None)`
matches choice_id to the first option with the same action_id as action.action_id. If no match, default to none

swagger_server/itm/itm_scenario_session.py
```diff
@@ -894,13 +905,13 @@ def get_available_actions(self, session_id: str, scenario_id: str) -> List[Actio
...
         
-        # TODO When an action takes place, that action should be removed from the probeYaml's option list!
-        # tackle one probe at a time, after all available actions are returned for that probe remove it from remaining probes
         if self.current_isso.probe_system.probe_yamls:
             for option in self.current_isso.probe_system.probe_yamls[self.current_isso.probe_system.current_probe_index].options:
                 if (not self.kdma_training):
                     option.assoc_action.pop('kdma_association', None)
                 actions.append(option.assoc_action)
+            # Always allow tagging a casualty
+            actions.append(Action(action_id="tag_action", scenario_id=scenario_id, action_type='TAG_CASUALTY', unstructured="Tag a casualty"))
```
dereknop, line 914: I might be misreading the code but it seems like this is only adding one option for tagging a casualty. Shouldn't we have one available tag action per casualty? If so we can just iterate through the casualty list and add one tag action corresponding to those casualty id's.

swagger_server/itm/itm_scenario_session.py
brianpippin: +1

swagger_server/itm/itm_scenario_session.py
brianpippin: the for loop I believe is adding the other actions, the last one just makes sure tagging is always added, is how I read it.

swagger_server/itm/itm_scenario_session.py
nextcen-dgemoets: This is a case where I'd choose readability over replacing a loop with one line, especially as I may be the one maintaining it.  It's part of a small function, so it's not helping to condense a large block of code, either.  Also, this wasn't code I added for the ticket.  I realize your approach is more Pythonesque.  Does @brianpippin  have an opinion or preference?

brianpippin: This is more a coding style choice and I don't want to be dictating anything there, so long as it is readable and understandable, I am good with either option.

brianpippin (approved):

PR #14, opened 2023-09-13
Itm 86: Add gradle config for swagger generation

KageRyu7:
Same branch as swagger but I renamed it to match the pattern used for other branches.

To test:
Manually delete the swagger_server/models folder
refresh your IDE or do a "git status" on the command line to see that git see the files as missing

run "./gradlew" to run the swagger generation

refresh your IDE or do a "git status" again. If all the models are generated correctly there should be no changes from gits perspective. The only file that is different in the init.py but its only unnecessary comments that are changed.


nextcen-dgemoets (changes requested):
Generation of `__init__.py` isn't working on Windows/Git Bash.  I'll re-review after that's addressed.  The rest of the stuff is minor.

nextcen-dgemoets (approved):

PR #16, opened 2023-09-14
ITM-96:  Add END_SCENARIO to the API and tweak custom September scenario/action progression

nextcen-dgemoets:
* Add END_SCENARIO to action space/API;
* SoarTech:  END_SCENARIO is always available; server only ends the scenario in response to END_SCENARIO ;
* BBN:  END_SCENARIO is not initially available; once ADM performs a MOVE_TO_EVAC action, the only available actions will be END_SCENARIO and TAG_PATIENT, the latter of which is always available.

Testing requires pulling ITM-96 branch from `itm-evaluation-client`.

### DIFF
--- a/swagger_server/itm/itm_scenario_session.py
+++ b/swagger_server/itm/itm_scenario_session.py
@@ -46,7 +46,7 @@ class ITMScenarioSession:
         self.first_answer: bool = True
         # hacky stuff for adept
         self.casualty_ids = []
-        self.probes_responded_to = []
+        self.adept_evac_happened = False
         # hacky thing for ST
         self.patients_treated = 0
         # adept or ST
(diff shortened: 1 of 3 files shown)


swagger_server/itm/itm_scenario_session.py
```diff
@@ -46,7 +46,7 @@ def __init__(self):
         self.first_answer: bool = True
         # hacky stuff for adept
         self.casualty_ids = []
-        self.probes_responded_to = []
+        self.adept_evac_happened = False
```
brianpippin, line 49: This is fine for the September deliverable, in the future, I think we want to name this to a generic variable name so we can use it for anything that might hold a scenario from having end_scenario available.

brianpippin (approved):

PR #22, opened 2023-09-21
ITM-98: Fix dictionary to bson conversion

brianpippin:


nextcen-dgemoets:
I have no actual way to test this, but these are both issues I'd flagged earlier based on code inspection when you raised the issue, so that inspection will serve as my "review".

PR #23, opened 2023-09-22
ITM-73: Support multiple active sessions

nextcen-dgemoets:
* Support multiple simultaneous sessions
* Re-use inactive sessions after a configurable amount of time

~~**Please note**:  there is some debugging and commented out lines that are included here to assist with debugging, but are intended to be removed after review.~~

Also, maybe there is a more Python-y way to implement `_reclaim_old_session()` and lines 92-98.

### DIFF
--- a/swagger_server/controllers/itm_ta2_eval_controller.py
+++ b/swagger_server/controllers/itm_ta2_eval_controller.py
@@ -8,9 +8,10 @@ from swagger_server.models.state import State  # noqa: E501
 
 from ..itm import ITMScenarioSession
 
-MAX_SESSIONS = 3     # Hard limit on simultaneous sessions
-#SESSION_TIMEOUT = 60 * 60 * 24  # 24 hour timeout in seconds
-SESSION_TIMEOUT = 25  # 10 second timeout in seconds
+MAX_SESSIONS = 5     # Hard limit on simultaneously active sessions
+SESSION_TIMEOUT = 60 * 60 * 24  # 24 hour timeout in seconds
+#MAX_SESSIONS = 3     # Hard limit on simultaneously active sessions
+#SESSION_TIMEOUT = 25  # 10 second timeout in seconds
 itm_sessions = {}     # one for each active adm_name
 session_mapping = {}  # maps session_id to adm_name and last active time
 """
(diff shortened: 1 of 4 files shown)


swagger_server/controllers/itm_ta2_eval_controller.py
```diff
@@ -1,19 +1,21 @@
...
 from swagger_server.models.state import State  # noqa: E501
-from swagger_server import util
 
 from ..itm import ITMScenarioSession
 
-ITM_SESSION = ITMScenarioSession()
+MAX_SESSIONS = 5     # Hard limit on simultaneously active sessions
+SESSION_TIMEOUT = 60 * 60 * 24  # 24 hour timeout in seconds
+#MAX_SESSIONS = 3     # Hard limit on simultaneously active sessions
+#SESSION_TIMEOUT = 25  # 10 second timeout in seconds
```
brianpippin, line 14: Do we need these commented out constants?  Look like Dupes.

swagger_server/controllers/itm_ta2_eval_controller.py
```diff
@@ -29,7 +31,13 @@ def get_alignment_target(session_id, scenario_id):  # noqa: E501
 
     :rtype: AlignmentTarget
     """
-    return ITM_SESSION.get_alignment_target(session_id=session_id, scenario_id=scenario_id)
+    session_dict = session_mapping.get(session_id)
+    adm_name = session_dict.get("adm_name") if session_dict else None
+    if not adm_name:
+        return 'Invalid Session ID', 400
+    else: # update access time
+        session_mapping[session_id] = {"adm_name": adm_name, "last_accessed": time.time()}
```
brianpippin, line 39: Lines 34-39 are duplicated 4 times in the code below, I am wondering if there was any way to make this into a function to reduce code duplication?

swagger_server/controllers/itm_ta2_eval_controller.py
nextcen-dgemoets: See the PR description/comment.  It was to help with debugging and will be removed, along with some logging.

swagger_server/controllers/itm_ta2_eval_controller.py
nextcen-dgemoets: Yep, I remember noticing that too, not seeing anything obvious/immediate, but wanting to revisit it.  I'll take a look.

nextcen-dgemoets: Thank you for having me revisit that-- I've pushed an update.

PR #25, opened 2023-10-05
ITM-106: Post-September demo cleanup tasks

nextcen-dgemoets:
Fixes a few things that were discovered after the final September release of the TA3 server and ADM client:
* Remove `patients_treated` from the server (and its use in `take_action`-- it’s a SoarTech scenario hack that ended up not being needed.
* Audit every description in the yaml files and ensure no reference to mvp, probes, and other MVP-1 phenomena.
* Improve error messages with invalid values, especially when validating action.
* Remove `EXPECTANT` from the relationship enum.
* Change lacerations of locations other than the thigh from `Hemostatic gauze` to `Pressure bandage`.
* Validate that the location and treatment parameter in `APPLY_TREATMENT` matches something in the YAML/API.
* Add ADM name to the `start_scenario` output, for ease of interpretation/parsing of output.
* `start_scenario` currently relies on an exception to indicate when the session is complete.  One of the problems with this is that it hides when there’s an actual exception, like communicating with TA1.
* Remove `scenario_id` from the `Action` class and from the `assoc_action` in the configured probe options.  It’s not needed, and ADMs should simply use the `scenario_id` they received from `start_scenario`.  Update the client appropriately.
  * Change the `id` of SoarTech’s scenario_1 to `st-september-2023-mvp2`.  This matches the `scenario_id` used by SoarTech, which now needs to match
* Sync the return codes between the YAML configuration and the server code.  Specifically, specifying an invalid `session_id` should result in a `400`, not a `404` return code.
* Remove unused  `test` session type.
* Text and fix if necessary the ability to take an optional `max_scenarios` in `start_session`.
  * Implement the disabled code for `start_scenario` to take an optional `scenario_id`.

### DIFF
--- a/swagger/swagger.yaml
+++ b/swagger/swagger.yaml
@@ -18,7 +18,7 @@ paths:
       - itm-ta2-eval
       summary: Start a new session
       description: Get unique session id for grouping answers from a collection of
-        scenarios/probes together
+        scenarios together
       operationId: start_session
       parameters:
       - name: adm_name
@@ -31,13 +31,13 @@ paths:
           type: string
       - name: session_type
         in: query
-        description: "the type of session to start (`test`, `eval`, or a TA1 name)"
+        description: "the type of session to start (`eval` or a TA1 name)"
         required: true
         style: form
         explode: true
         schema:
           type: string
-          example: test
+          example: eval
       - name: kdma_training
         in: query
         description: whether or not this is a training session with TA2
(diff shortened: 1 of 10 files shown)


brianpippin (approved):

swagger/swagger.yaml
```diff
@@ -49,8 +49,8 @@ paths:
           default: false
       - name: max_scenarios
         in: query
-        description: "the maximum number of scenarios requested, supported only in\
-          \ `test` sessions"
+        description: "the maximum number of scenarios requested, not supported in\
```
KageRyu7, line 52: The previous description shows that this was only used for 'test' sessions (Not eval or ta1) but they were removed. By saying that it's not supported in 'eval' sessions does this mean it is supported in ta1? Or does is this just no longer supported

nextcen-dgemoets: Correct-- it is supported in TA1 (`soartech` & `adept`) sessions, but not `eval` sessions.  Basically I removed `test`, but made sure that `max_scenarios` works in `soartech` & `adept`.

PR #26, opened 2023-10-20
ITM-111: Break-up itm_scenario_session into different files

nextcen-dgemoets:
* Moved action validation/processing code to `ITMActionHandler`
* Refactored some action handling code
* Decoupled history JSON output from DB, and only init DB once, 
* Moved history to `ITMHistory` with its own ability to write JSON output
* After discussion, consume supplies (and advance simulated time) even if treatment doesn't treat the injury
* Removed `git_push.sh` (for ITM-93)

We should be able to break out the scenario/probe advancement code into another module down the road, but since so much of this code will be changing in the coming months, it seemed best to wait until this area is largely re-written.


brianpippin (approved):
Nothing stood out to me, I think the new itm_history is on the right path.

swagger_server/itm/itm_action_handler.py
```diff
@@ -0,0 +1,416 @@
...
+
+    def set_isso(self, isso):
+        self.current_isso = isso
+
+
+    def _reveal_injuries(self, source: Casualty, target: Casualty):
+        if target.visited: # Don't reveal injuries in visited casualties
+            return
+
+        for source_injury in source.injuries:
```
dereknop, line 31: Totally up to preference and whatever seems more readable/maintainable but here's a suggestion for avoiding having a nested for loop here 
`def should_reveal_injury(source_injury, target):
        return source_injury.name in hidden_injury_types and \
               not any(target_injury.name in hidden_injury_types for target_injury in target.injuries)`
`revealed_injuries = [source_injury for source_injury in source.injuries if should_reveal_injury(source_injury, target)]
  target.injuries.extend(revealed_injuries)`

dereknop, line 31: If you are interested in not having a nested for loop here we could split this into a separate function like this.

```python
def should_reveal_injury(source_injury, target):
        return source_injury.name in hidden_injury_types and \
               not any(target_injury.name in hidden_injury_types for target_injury in target.injuries)
```

And then call it inside _reveal_injuries like this:

```python
revealed_injuries = [source_injury for source_injury in source.injuries if should_reveal_injury(source_injury, target)]
target.injuries.extend(revealed_injuries)
```

dereknop (approved):

swagger_server/itm/itm_history.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+    Class for managing ADM action history.
+    """
+
+    def __init__(self):
+        """
+        Initialize an instance of ITMSupplies.
+        """
+        self.history = []
+        self.filepath = 'itm_db_local_output/'
```
lawerencem, line 15: Is the trailing '/' necessary?  It's an os specific path separator.  It may not matter since we use os.path.join on it later and that has some fancy join logic.  Otherwise, it should be os.sep.

swagger_server/itm/itm_action_handler.py
```diff
@@ -0,0 +1,416 @@
...
+                self.session.history.add_history(
+                    "Check Pulse",
+                    {"Session ID": self.session.session_id, "Casualty ID": casualty.id},
+                    casualty.vitals.hrpmin)
+                return self.times_dict['CHECK_PULSE']
+
+
+    def check_respiration(self, casualty: Casualty):
+        """
+        Check the respiration of the specified casualty in the scenario.
```
lawerencem, line 271: check_pulse and check_respiration are nearly identical and share a lot of code with check_all_vitals.  If we add any other checks, we might consider refactoring these to pull out all the common code or have a generic check_vitals that takes an additional array of things to check (e.g., ['PULSE', 'RESPIRATION']) and sets any item in the array in addition to the basic checks of breathing, consciousness, and mental status.

swagger_server/itm/itm_history.py
nextcen-dgemoets: It turns out the `/` is necessary.  Changed to `os.sep`.

swagger_server/itm/itm_action_handler.py
nextcen-dgemoets: Agreed.  We'll likely be adding blood ox level, so that change is probably in our future.

swagger_server/itm/itm_action_handler.py
nextcen-dgemoets: I think ultimately the "hiddenness" of injuries will be a property of the injury and not the scenario, so this code will need to be revisited.  I made your change in the meantime.

nextcen-dgemoets:
@brianpippin  @lawerencem and @dereknop :
In working on ITM-114 (human adm sim), I noticed a potential problem on the server-side.  We added END_SCENARIO to the action space and that was supposed to be the only way to end a scenario.  However, they could still request a new scenario via start_scenario the way they used to.  Good ADMs won't do that, but it needed to be handled.

So I added some code in this last commit to handle "aborting" a scenario if the ADM calls start_scenario without first explicitly ending it via END_SCENARIO action.  When you abort, the necessary clean-up happens, but nothing gets saved to the DB for that scenario.  Alternatively, I could just return a Bad Request / 400 if they call start_scenario if there's an existing scenario that hasn't already been ended.  Thoughts?

lawerencem:
So long as they are functionally equivalent and ending a scenario is handled the same way in both cases, I don't see a major issue with it now.  

However, as the scenarios become more complex, there's the chance that someone may inject new end logic and fail to do it in both spots.  Also, doing clean-up as a silent side effect is generally a bad idea.

My preference would be to return an error as you said and make them follow correct procedure.  Cleaning up things for them becomes a bad habit.  If we truly don't trust them to do that though, at least explicitly document that "start" is really "end + start"

nextcen-dgemoets:
> So long as they are functionally equivalent and ending a scenario is handled the same way in both cases, I don't see a major issue with it now.
> 
> However, as the scenarios become more complex, there's the chance that someone may inject new end logic and fail to do it in both spots. Also, doing clean-up as a silent side effect is generally a bad idea.
> 
> My preference would be to return an error as you said and make them follow correct procedure. Cleaning up things for them becomes a bad habit. If we truly don't trust them to do that though, at least explicitly document that "start" is really "end + start"

Well...they're not functionally equivalent.  If they END_SCENARIO then start a scenario, then the previous scenario is saved to history which is written to disk/database.  If they start a new scenario without ending the previous one, then the old scenario is aborted and not saved/written.

The logic is only written in one place-- there's just a flag on whether it's a "normal" ending of a scenario or an abort.  The server logs that it aborted the previous scenario, but the clean-up is necessary so that the new scenario can start with a clean slate.  It does not report to the ADM that the scenario was aborted, so in that sense it's silent.

I'm not opposed to removing the abort code and just returning a bad request-- hence the original question.  It seems useful to the human ADM sim, but really that's mostly just a test tool.  I could also do nothing for this PR, and write a ticket where we determine the right approach and implement it.  Let's see what @brianpippin and/or others think.

nextcen-dgemoets:
I decided to back out the scenario abortion code (handling when ADMs call `start_scenario` before calling `END_SCENARIO`) in favor of a new ticket, [ITM-121](https://nextcentury.atlassian.net/browse/ITM-121).  That ticket also contains a couple other minor bugs, but none of those changes have anything directly to do with this ticket.

PR #28, opened 2023-11-02
ITM-124 - server edition

kdrumm:
- copied over new swagger.yaml.  
- Added function to get list of valid strings for enums.  Now enums should be tied to swagger.yaml and not listed in multiple locations
- Removed some magic strings


swagger_server/util.py
```diff
@@ -4,6 +4,8 @@
 import typing
 from swagger_server import type_util
 
+def get_swagger_class_enum_values(klass):
+    return [getattr(klass,i) for i in dir(klass) if not i.startswith("_") and isinstance(getattr(klass,i), str)]
```
nextcen-dgemoets, line 8: This file was generated by swagger.  That said, I don't expect we'll re-generate the entire server from swagger, and our gradle target doesn't re-generate it.  I think it's okay then, right?

nextcen-dgemoets (changes requested):
Looks good.

I think my only real comments are the same ones from the client version regarding the YAML (e.g., new enums, removing `TriageCategory`, etc.).

kdrumm:
I can run locally, but I can't run from docker.  I get the following error:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/local/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/usr/src/app/swagger_server/__main__.py", line 6, in <module>
    from swagger_server import encoder
  File "/usr/src/app/swagger_server/encoder.py", line 1, in <module>
    from connexion.apps.flask_app import FlaskJSONEncoder
ModuleNotFoundError: No module named 'connexion.apps.flask_app'
```

It used to work so this might need to be fixed prior to any merge.

kdrumm:
i cleared my virtual environment and i get the same error locally.  This appears to be due to a new swagger-ui-bundle version 1.1.0 released on Wednesday.  If I set the requirements.txt to `swagger-ui-bundle == 0.0.9`, everything works fine.  Do we want to pin the version or should we try adding flask to the requirements.txt?

kdrumm:
Writing this out so others don't have to dive as deep as I did.  The core of the issue is that upgrading from `swagger-ui-bundle:0.0.9` to `swagger-ui-bundle:1.1.0` results in an error `ModuleNotFoundError: No module named 'connexion.apps.flask_app'`.  The version increment is weird, but those are adjacent versions.

The error is really caused by swagger switching to a different `connexion` version `swagger-ui-bundle:0.0.9` uses `connexion:2.14.2` which requires `flask` and `swagger-ui-bundle:1.1.0` uses `connexion:3.0.0` (released yesterday!) which does not require `flask` directly.    `connexion:3.0.0` does appear to use flask based on their changelog, but I'm not sure if it pulls it in or not through dependencies.

Regardless, I think the upgraded `connexion` is a major version change and seems to have made API changes.   Detailed changes can be see [here](https://connexion.readthedocs.io/en/stable/v3.html).  Also, [changelog](https://github.com/spec-first/connexion/releases/tag/3.0.0)

the new `FlaskApp` from `connexion` does not include the object `FlaskJSONEncoder` which is the DIRECT cause of the error.

My recommendation:

- Pin the versions to the previous versions for now
- As a separate ticket, evaluate usefulness and perform an upgrade.

nextcen-dgemoets:
I'm fine with your recommendation.  Does @brianpippin  want to weigh in on this?

Also, like in client, I believe `swagger_client/models/triage_category.py` needs to be manually removed and the gradle task re-run to remove references to it.

kdrumm:
Created a ticket here: https://nextcentury.atlassian.net/browse/ITM-129

nextcen-dgemoets (approved):

brianpippin (approved):

PR #31, opened 2023-11-29
ITM-145: Globally replace 'casualty' with 'character'

nextcen-dgemoets:
Changes made to Swagger YAML, server code, and other configuration.


swagger_server/itm/itm_scenario_configs/adept/scenario_1/probe_1.yaml
lawerencem, line 1: A quick text search for "casual" will find 6 values that have "casualty-" in them as well as one id, "select-casualty-abc."

lawerencem: It's for the probe "Who do you treat first between casualties A, B and C?"

nextcen-dgemoets: Yes, those are authored by TA1 (including the IDs) so I have no control over them.

brianpippin (approved):

PR #33, opened 2023-12-07
ITM-150:  In training mode, let TA1 session span multiple scenarios

nextcen-dgemoets:
Training sessions can span multiple scenarios.  To support this, moved handle to TA1 controller to session level so it can (in training mode) persist across scenarios.

Note that this ticket does NOT support a mix of TA1 scenarios in a training session.  If there's a need for a training session to include a mix of soartech and adept scenarios, then we probably need a new session type, e.g., `training`, and a way to configure the scenarios therein.

To test:  run both the bbn/adept and soartech TA1 servers, and run our clients with training mode toggled off, then toggled on, and ensure that the ADM server is creating new TA1 sessions at the appropriate times (e.g., once at session start when in training mode, and after every scenario when not in training mode).  Make sure `ta1_integration` is set to `True` in the server. 
 Remember that you can specify a number of scenarios to run as part of the `--session` switch.  Here are a few things to try:
-  `python itm_minimal_runner.py --adm_name my_adm1 --session soartech 3 --kdma_training`
-  `python itm_minimal_runner.py --adm_name my_adm2 --session soartech 3`
-  `python itm_minimal_runner.py --adm_name my_adm3 --session adept 3 --kdma_training`
-  `python itm_minimal_runner.py --adm_name my_adm3 --session adept 3`
-  `python itm_minimal_runner.py --adm_name my_adm  --eval`


kaitlyn-sharo (commented):
I don't feel qualified to approve this since this is the first time I'm digging deeper into this code, but the changes look good to me and it seemed to run with soartech. I am having some problems getting adept to run and don't want to spend too much unsupervised time on it since I'm nearing my hourly limit for ITM.

lawerencem (approved):

dereknop (approved):
Tested it using the development branch of the client and everything worked for me!

PR #39, opened 2024-02-09
ITM-210: Get Metrics Eval server ready for TA2

nextcen-dgemoets:
* Updated YAML and re-generate model files based on latest decisions;
* Merged scene state into scenario state;
* Added condition-handling needed for the metrics eval (probe and actions), plus two other simple conditionals;
  * Added condition semantics (and, or, not)
* Added some logging / history;
* Added reusable supplies, treatable burns, and some new injuries;
* Fixed ending scenes/scenarios;
* Updated `get_available_actions` to include unmapped actions that aren't `restricted`.
* Consolidated scene state, which will enable characters to persist across scenes like they did in MVP; and
* Added ADEPT and SoarTech training and evaluation scenario configuration files.

To test, be sure to pull the latest from the [client's ITM-210 branch](https://github.com/NextCenturyCorporation/itm-evaluation-client/tree/ITM-210).  I've left in some debugging in the server that will assist in testing; it should be removed before this PR is merged.  I will continue to test as well, as time permits, and after this PR is merged.

To run all scenarios, use:
* `python itm_minimal_runner.py --adm_name my_adm --session adept`
* `python itm_minimal_runner.py --adm_name my_adm --session adept --kdma_training`
* `python itm_minimal_runner.py --adm_name my_adm --session soartech`
* `python itm_minimal_runner.py --adm_name my_adm --session soartech --kdma_training`


swagger_server/itm/itm_eval_scenarios/adept/scenario_1/alignment_target.yaml
brianpippin, line 1: I think we want to differentiate that these scenarios are more for the metrics evaluation and not just evaluation scenarios.  I could see a scenario where we reuse these for something else.

swagger_server/itm/itm_scenario.py
```diff
@@ -1,73 +1,240 @@
...
+            current_state.unstructured = target_state.unstructured
+
+        # Threat state
+        if target_state.threat_state:
+            target = target_state.threat_state
+            current = current_state.threat_state
+            if not current:
+                current = target
+            else:
+                current.unstructured=target.unstructured if target.unstructured else current.unstructured
```
brianpippin, line 182: I know this is probably something a linter would handle if we had one, but the spaces not being around the "=" caught my eye.  Totally nitpicky.

dereknop:
The logging you added helps follow along. Running all of the scenarios with the client worked with no errors. Is it expected behavior that I am getting "Got session alignment score 0.0 from TA1" each time I run a scenario with --kdma-training?

swagger_server/itm/itm_scenario.py
```diff
@@ -1,73 +1,240 @@
...
+                current.threats = target.threats if target.threats else current.threats
+            current_state.threat_state = current
+
+        # Mission
+        if target_state.mission:
+            target = target_state.mission
+            current = current_state.mission
+            if not current:
+                current = target
+            else:
```
dereknop, line 192: I'm not sure if this list of attributes will continue to grow or be modified in the future but here is a way to maybe make this section a little more maintainable
```
attributes = [
        "unstructured", "mission_type", "character_importance", 
        "civilian_presence", "communication_capability", "medical_policies", 
        "political_climate", "roe"
    ]
   
    for attr in attributes:
        setattr(current, attr, getattr(target, attr) or getattr(current, attr))
    
current_state.mission = current
```
 This way you can just maintain a dictionary of all of the attributes to loop over instead of having to manually maintain this section.

dereknop: The same idea (if feasible here) applies to sim env and decision env below

nextcen-dgemoets:
> The logging you added helps follow along. Running all of the scenarios with the client worked with no errors. Is it expected behavior that I am getting "Got session alignment score 0.0 from TA1" each time I run a scenario with --kdma-training?

Yes-- because we're not connecting to a TA1 server, it logs the session alignment as zero.

swagger_server/itm/itm_scenario.py
nextcen-dgemoets: I like nit-picky.   ;)  Will fix in next commit.

swagger_server/itm/itm_eval_scenarios/adept/scenario_1/alignment_target.yaml
nextcen-dgemoets: I haven't really updated the alignment targets yet because they haven't been sent from the TA1s...but let's talk about differentiating between metrics eval and real eval.  Usually we just update the scenario files (incl. alignment targets) at each iteration (and tag the repo).

swagger_server/itm/itm_scenario.py
nextcen-dgemoets: This looks really promising!  I knew there would be a slick Python-y way to do this, but I just needed to get something working out there.  :)  To add to your solution, I think we can use they keys from `Mission.attribute_map` (for example) that won't need any maintenance.  I'll try it out and test both ways and ensure the same results.

nextcen-dgemoets: As discussed in Teams, your solution work, as did my small addition mentioned above (except that I had to use an instance of the model object since `attribute_map` isn't static.

And once the changes were made, it became clear we could refactor it further.  Thank you!

swagger_server/itm/itm_eval_scenarios/adept/scenario_1/alignment_target.yaml
nextcen-dgemoets: We discussed that in the future, we will have the metrics eval and real eval scenario files to exist side-by-side in the repo as opposed to relying on git releases.  It's something we can add between evals.

brianpippin (approved):

dereknop (approved):

PR #42, opened 2024-02-29
ITM-241: Support pre-configured treated injuries and initially visible vitals

nextcen-dgemoets:
This PR includes:
1. Support pre-configured treated injuries and initially visible vitals;
2. Fix scene transitioning bug;
3. Add missing action times (elapsed time associated with ADM actions); and
4. Made minor tweak to DIRECT_MOBILE_CHARACTERS (limit to to CALM or UPSET characters)

To test, use a custom scenario crafted to test the changes, attached [here](https://github.com/NextCenturyCorporation/itm-evaluation-server/files/14502186/scenario.zip).  Place it in `itm/itm_eval_scenarios/adept/scenario_1/scenario.yaml`, replacing the `scenario.yaml` that's currently there.

Also, use the [client's ITM-241 branch](https://github.com/NextCenturyCorporation/itm-evaluation-client/tree/ITM-241) and run: 

`python itm_minimal_runner.py --adm_name foobar --session adept --scenario test-1`.

See the comment added below for an explanation of the results and how it tests (1) and (2) above.

Feel free to change the scenario and/or [action path](https://github.com/NextCenturyCorporation/itm-evaluation-client/blob/ITM-241/swagger_client/config/adept_action_path.json) to test further.

Also, I'm open to discussion on how I handled treating treated injuries.  I made it so that it doesn't take as most regular treatments (not that elapsed time matters in the Metrics Eval), but that vitals aren't discovered either.

### DIFF
--- a/swagger_server/itm/itm_action_handler.py
+++ b/swagger_server/itm/itm_action_handler.py
@@ -303,7 +303,8 @@ class ITMActionHandler:
         for character in self.session.state.characters:
             for isd_character in self.current_scene.state.characters:
                 if isd_character.id == character.id:
-                    if isd_character.vitals.ambulatory:
+                    if isd_character.vitals.ambulatory and \
+                    isd_character.vitals.mental_status in [MentalStatusEnum.CALM, MentalStatusEnum.UPSET]:
                         character.vitals.ambulatory = True
                         character.vitals.conscious = True
         return self.times_dict["DIRECT_MOBILE_CHARACTERS"]
--- a/swagger_server/itm/itm_action_handler.py
+++ b/swagger_server/itm/itm_action_handler.py
@@ -34,10 +34,10 @@ class ITMActionHandler:
         self.current_scene = scenario.isd.current_scene
 
     def _reveal_injuries(self, source: Character, target: Character):
-        if target.visited: # Don't reveal injuries in visited characters
-            return
+        if target.visited:
+            pass # Character could be pre-configured visited but with discoverable injuries
 
-        # Add discoverable injuries target character (with discovered status).
+        # Add discoverable injuries to target character (with discovered status).
         revealed_injuries = [source_injury for source_injury in source.injuries if source_injury.status == InjuryStatusEnum.DISCOVERABLE]
         for injury in revealed_injuries:
             injury.status = InjuryStatusEnum.DISCOVERED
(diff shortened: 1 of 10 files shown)


nextcen-dgemoets:
For reviewers: when you run the client, this is what you should see:
- In the opening state displayed by the client, `Mike` has full vitals because his `visited` is true, but he has no injuries because his injuries are `discoverable`, not `visible` or `treated`.
- After the `CHECK_ALL_VITALS` action on `Mike, the previously invisible `discoverable` injury is now visible and `discovered`.
- After the `SITREP` action on `Civilian_01`, vitals are still hidden except for his `mental_status` of `UNRESPONSIVE` (because he didn't respond to the sitrep to reveal basic vitals or injuries.
- After the `TAG_CHARACTER` on `Mike`, you get a whole new cast of characters because the scene changed.  `Captain_01` (Katie) has a pre- `treated` burn injury, but no displayed vitals because her `visited` is False.  Elapsed time is 50 (seconds).
- After the `APPLY_TREATMENT` to `Captain_01` 's already treated injury reveals no additional information, but the elapsed time only increased 10s to 60, instead of a full 60s if the injury was actually treated.  However, her vitals remain hidden.
- After the `SEARCH`, the scenario ends without sending a probe response, because the scene is configured to transition to the next scene after the probe **OR** after 300s have elapsed in the scenario (note the final elapsed time).  This shows that the scene transitioning bug is fixed, because `SEARCH` was not an action mapping.

brianpippin (approved):
Code looked good and nothing caught my eye running.

phile-caci (approved):
I think every thing looks ok... I updated the scenario.yaml, and ran 
`python itm_minimal_runner.py --adm_name foobar --session adept --scenario test-1.`
on the client using the ITM-241 branch, played around with the action path, didn't see any issues.  

QQ: On the action path ran different runs toggling tag-mike-immediate to tag-mike-minimal.  In the runner output the tag value for mike is always `'tag': None,` I assuming that's right?? but wanted to make sure given my extremely limited knowledge on how this should work.

nextcen-dgemoets:
> QQ: On the action path ran different runs toggling tag-mike-immediate to tag-mike-minimal. In the runner output the tag value for mike is always `'tag': None,` I assuming that's right?? but wanted to make sure given my extremely limited knowledge on how this should work.

Mike actually shouldn't be in the state after the TAG_CHARACTER action, because that action triggered a scene change.  The new characters (Bob and Katie) are untagged.

If you want to see tagging work (which I just did to make sure, hehe) is switch the order of the tagging action and, say, action4 in your action paths.

PR #44, opened 2024-03-13
ITM-252: Better output file naming for JSON

nextcen-dgemoets:
Change history filename to `<scenario_id>-<ta1>-<alignment>-<date>.json` e.g.
- e.g., `MetricsEval.MD4-Jungle-adept-high-Mar13-11.45.36.json`
- e.g., `submarine-1-soartech-high-Mar13-11.44.45.json`

To test, run the development client with the  `--eval` flag and check out the results in the server's `itm_history_output` folder.
This requires both TA1 servers running, but you can also try a standalone session (e.g., with `--session adept` instead of `--eval`) but change `self.save_history = True` in `itm_session.py`.

### DIFF
--- a/swagger_server/itm/itm_history.py
+++ b/swagger_server/itm/itm_history.py
@@ -40,7 +40,7 @@ class ITMHistory:
         self.history.append(history_to_add)
 
 
-    def write_to_json_file(self) -> None:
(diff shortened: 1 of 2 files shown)


phile-caci (approved):
Looks good to me
image

brianpippin (approved):

PR #46, opened 2024-03-13
ITM-253 - Eval server history to s3

phile-caci:
1. Get a temporary aws access token: 
  a. you'll need aws-cli installed
  b. grab [saml-caci-all.zip](https://caci.servicenowservices.com/sys_attachment.do?sys_id=1c127be01b34f5d01afb0e9ee54bcbfa) from [here](https://caci.servicenowservices.com/caci?id=kb_article_view&table=kb_knowledge&sys_kb_id=4c12b3e01b34f5d01afb0e9ee54bcb9b&recordUrl=%2Fkb_view.do%3Fsys_kb_id%3D4c12b3e01b34f5d01afb0e9ee54bcb9b) (2nd section - # 3)
  c. unzip the above and run your OS specific command
  d. it'll generate a config and credentials file
  e. in your .aws folder (you'll have one in your home dir after installing the aws-cli), back up your config/crendetials file and replace it with the ones the saml tool created
  f. in the credentials file rename `[saml]` to `[default]`
  g. you should now be able to issue aws commands, try `aws s3 ls`, this will list all our buckets in s3
3. Start a TA1 Server (i.e. Adept) - https://gitlab.com/itm-ta1-adept-shared/adept_server
4. In itm-evaluation-server, Change line 58 `self.save_history` = False to `self.save_history = True`
5. In itm-evaluation-client, run something, i.e.:  `python itm_minimal_runner.py --adm_name test_misaligned --session adept`
6. History should be in: `itm-evaluation-server/itm_history_output`
7. Check s3 bucket: `aws s3 ls itm-ui-assets`, the history files should be there.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,11 +1,6 @@
-<<<<<<< HEAD
 boto3 >= 1.34.61
-connexion >= 2.6.0
-connexion[swagger-ui] >= 2.6.0
-=======
 connexion >= 2.14.0, < 3.0
 connexion[swagger-ui] >= 2.14.0, < 3.0
->>>>>>> development
 python_dateutil == 2.6.0
 setuptools >= 21.0.0
 swagger-ui-bundle == 0.0.9
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-boto >= 1.34.61
+boto3 >= 1.34.61
 connexion >= 2.6.0
 connexion[swagger-ui] >= 2.6.0
 python_dateutil == 2.6.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+boto >= 1.34.61
 connexion >= 2.6.0
 connexion[swagger-ui] >= 2.6.0
 python_dateutil == 2.6.0
--- a/swagger_server/itm/itm_history.py
+++ b/swagger_server/itm/itm_history.py
@@ -65,6 +65,7 @@ class ITMHistory:
         with open(full_filepath, 'w') as file:
             # Convert Python dictionary to JSON and write to file
             json.dump({'history': self.history}, file, indent=2)
+
         self.save_json_to_s3(os.getcwd() + os.path.sep + full_filepath, filebasename)
 
     def save_json_to_s3(self, full_filepath, file_name) -> bool:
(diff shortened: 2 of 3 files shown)


brianpippin:
> @brianpippin Do you want this PR to go to `development` or straight to `main`? Are we still using `development` the same way as before?

Development then Main so we can test the Jenkins job that Phi has for auto deploying main when pushed.

nextcen-dgemoets (commented):
I got almost everything to work-- files were written locally and to S3.  I was not able to retrieve the files from S3 (Bad Request) to compare to the JSON.

One question:  why are the files copied to S3 without the `.json` extension?

phile-caci:
> I got almost everything to work-- files were written locally and to S3. I was not able to retrieve the files from S3 (Bad Request) to compare to the JSON.
> 
> One question: why are the files copied to S3 without the `.json` extension?

Regarding the json extension, that's odd, they were there yesterday.... checking

phile-caci:
@nextcen-dgemoets , I missed the change that effected the filebasename, it's corrected now.  The json extension should be there.

There seems to be something up with our federated AWS accounts, I can't even `aws s3 ls` now...  I'll give it a bit to see if it comes back

nextcen-dgemoets (approved):

nextcen-dgemoets:
@phile-caci ,   Matt from SoarTech has asked to have a toggle to disable writing to S3 in `eval` mode (something like the current `save_history` toggle.  I had considered suggesting the same thing.  I can add it if you like, or you're welcome to.

phile-caci:
Not a problem I can add that.

phile-caci:
@nextcen-dgemoets , the flag is added.

nextcen-dgemoets:
@phile-caci   I hope you don't mind, but I set saving history back to False by default, but true for `eval` sessions, where saving to S3 is also enabled.  Also, I just logged S3 errors instead of aborting.  These changes are in response to TA2s wanting to run `eval` sessions prior to the evaluation.

phile-caci:
> @phile-caci I hope you don't mind, but I set saving history back to False by default, but true for `eval` sessions, where saving to S3 is also enabled. Also, I just logged S3 errors instead of aborting. These changes are in response to TA2s wanting to run `eval` sessions prior to the evaluation.

Thanks @nextcen-dgemoets!  Just rested when session was eval, history was uploaded. When it wasn't --eval, it did not upload.

brianpippin (approved):

PR #48, opened 2024-03-18
Development to Main

brianpippin:


nextcen-dgemoets (approved):
Looks good.  I wasn't connected to S3, but the history looks good and got the (green) error message for not copying to S3.

phile-caci (approved):
Tested with and without eval, and against soretech and adept, with eval files did go to s3.
didn't see any issues.

PR #52, opened 2024-03-19
ITM-233: Support persistent characters

nextcen-dgemoets:
* Added `persist_characters` to `Scene` model object and implement in server
* Tweaked behavior of `DIRECT_MOBILE_CHARACTERS` to be more realistic
* Added temporary/test scenario to assist reviewers

**This ticket will only persist characters, with the limitation that there’s no way to *remove* a character without replacing the entire cast of characters (which you do for a narrative scene change or action interruption by setting `persist_characters` to false and specifying characters).  If `persist_characters` is true, any specified characters will be added to the state, replacing those with the same `character_id`.**

Use with [ITM-233 feature branch](https://github.com/NextCenturyCorporation/itm-evaluation-client/tree/ITM-233) in the client.  See testing instructions in the [client pull request](https://github.com/NextCenturyCorporation/itm-evaluation-client/pull/27).


dereknop (approved):

kaitlyn-sharo (approved):
Worked for me, code looks good!

PR #55, opened 2024-04-04
ITM-329:  Fix TA1 session contamination when running multiple ADMs simultaneously

nextcen-dgemoets:
Basically, we just need to clone the TA1 controllers so that each scenario has its own TA1 session ID.  Duh.
However, I added some logging and config to make this easier to test.  This will be removed prior to merge.

To test:
- Run this branch of the server and have three itm-scenario-client windows open (`development` branch is fine).
- Have your favorite TA1 server running (I used ADEPT because it doesn't require Docker).
- Start the first client:  `python itm_minimal_runner.py --adm_name adm1 --session adept`
- Once the first scenario is complete, start the second client with an `adm_name` of adm2.
- Once the first scenario is complete in the second client, start the third client with `adm_name` of adm3.  There should soon be three clients running simultaneously.
- Ensure that there is no green logging in the server telling about contamination is session or probe alignment.

**NOTE**: if your machine is very fast and the sessions end too quickly, then just start all three clients one after another as fast as possible.

To test thoroughly:
- Go to `itm_history_output`, and run grep commands like the following:
  - `grep -i ubmari *rban* *esert* *ungle*`
  - `grep -i esert *rban* *ubmari* *ungle*`
  - `grep -i ungle *rban* *ubmari* *esert*`
  - This should ensure that there are no e.g. submarine scenario IDs in non-submarine scenarios.
- Remove the `deepcopy` from `itm_session.py` line 453 and repeat the initial test.
  - Ensure that you DO see some green contamination logging.
  - Try those greps and get alarming results.


kaitlyn-sharo (approved):
Everything works as described

dereknop (approved):
I got the expected behavior on everything (no contamination until I removed the deepcopy), and a reflexes workout switching from the three different consoles fast enough to have them running simultaneously.

PR #59, opened 2024-04-16
ITM-300: add adm_profile optional session parameter

nextcen-dgemoets:
Currently, just logs the profile to history.  A new ticket would expand the notion of ADM profile for multiple KDMAs, which will probably entail some discussion down the line.

Test using the corresponding [ITM-300](https://github.com/NextCenturyCorporation/itm-evaluation-client/tree/ITM-300) client branch.

[ITM-300]: https://nextcentury.atlassian.net/browse/ITM-300?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/swagger_server/controllers/itm_ta2_eval_controller.py
+++ b/swagger_server/controllers/itm_ta2_eval_controller.py
@@ -115,7 +115,7 @@ def start_session(adm_name, session_type, adm_profile=None, kdma_training=False,
 
     :param adm_name: A self-assigned ADM name.
     :type adm_name: str
-    :param session_type: the type of session to start `test`, `eval`, or a TA1 name)
+    :param session_type: the type of session to start (`test`, `eval`, or a TA1 name)
     :type session_type: str
     :param adm_profile: a profile of the ADM in terms of its alignment strategy
     :type adm_profile: str
--- a/swagger_server/controllers/itm_ta2_eval_controller.py
+++ b/swagger_server/controllers/itm_ta2_eval_controller.py
@@ -108,18 +108,20 @@ def _reclaim_old_session():
             session_mapping.pop(session_id)             # From both places
             return ITMSession()
 
-def start_session(adm_name, session_type, kdma_training=None, max_scenarios=None):  # noqa: E501
+def start_session(adm_name, session_type, adm_profile=None, kdma_training=False, max_scenarios=None):  # noqa: E501
     """Start a new session
 
     Get unique session id for grouping answers from a collection of scenarios/probes together # noqa: E501
 
-    :param adm_name: A self-assigned ADM name.  Can add authentication later.
+    :param adm_name: A self-assigned ADM name.
     :type adm_name: str
-    :param session_type: the type of session to start (&#x60;test&#x60;, &#x60;eval&#x60;, or a TA1 name)
+    :param session_type: the type of session to start `test`, `eval`, or a TA1 name)
     :type session_type: str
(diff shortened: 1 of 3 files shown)


dereknop (approved):

swagger_server/controllers/itm_ta2_eval_controller.py
```diff
@@ -108,18 +108,20 @@ def _reclaim_old_session():
...
+def start_session(adm_name, session_type, adm_profile=None, kdma_training=False, max_scenarios=None):  # noqa: E501
     """Start a new session
 
     Get unique session id for grouping answers from a collection of scenarios/probes together # noqa: E501
 
-    :param adm_name: A self-assigned ADM name.  Can add authentication later.
+    :param adm_name: A self-assigned ADM name.
     :type adm_name: str
-    :param session_type: the type of session to start (&#x60;test&#x60;, &#x60;eval&#x60;, or a TA1 name)
+    :param session_type: the type of session to start `test`, `eval`, or a TA1 name)
```
kaitlyn-sharo, line 118: Really small thing, but it looks like a parenthesis was lost here

swagger_server/controllers/itm_ta2_eval_controller.py
```diff
@@ -149,6 +151,7 @@ def start_session(adm_name, session_type, kdma_training=None, max_scenarios=None
         adm_name=adm_name,
         session_type=session_type,
         kdma_training=kdma_training,
+        adm_profile=adm_profile if adm_profile != 'None' else None,
```
kaitlyn-sharo, line 154: I haven't tested yet, but just looking at the code in server and client, I am curious how adm_profile could ever be the string "None". I see it as an empty string and as the value None. Is there some place I'm missing where it is a string None, and that will always be passed in if it's an empty string or the value None?

kaitlyn-sharo (approved):
Works for me, just want to make sure the 'None' in a string is intended. It probably is, but caught my eye

swagger_server/controllers/itm_ta2_eval_controller.py
nextcen-dgemoets: It's been a while since I added this / was testing this, but I seem to recall that when I was using the `itm_minimal_runner` client and did not supply an `adm_profile` parameter on the command line, that it would get propagated to the TA2 controller as the string 'None'.  I believe this is because the client always sends the `adm_profile` parameter, even if it wasn't on the command line.  Presumably it sends `None` across the wire, but the server code interprets it as the string "None".

You can test this by using the `itm_minimal_runner` client, don't set `adm_profile`, and see what the TA2 controller gets.  I *believe* it will get the string 'None', hence the extra code.

Another approach could have been to change the client only to send the `adm_profile` if a value was set.  But adding the code to the server accounts for if the client doesn't do this, which it arguably should.

swagger_server/controllers/itm_ta2_eval_controller.py
nextcen-dgemoets: I will fix this in my upcoming PR for ITM-227, thanks.

swagger_server/controllers/itm_ta2_eval_controller.py
kaitlyn-sharo: That makes sense. I was wondering if it was some conversion "problem" from client to server.

PR #60, opened 2024-04-22
ITM-227: Implement remaining configurable transition conditions

nextcen-dgemoets:
Implemented the following transition conditions:
- a given list of probes responses has been sent;
- a given supply (or supplies) reaches a threshold; and
- a given character's vitals meets criteria (e.g., heart_rate = NONE, breathing = NONE)

Also simplified the code for the overall determination of whether conditions are met.

**NOTES**:
- I've left in some logging to help with supplies and vitals.  Should be removed prior to merge.
- Note that enum vitals will only trigger if the value is exact.  In other words, if the condition is "breathing = FAINT", this will not be triggered if breathing is NONE (even though we as humans know that no breathing is worse than faint breathing).  There is no order or ordinal value for the enums.  If you want to define a condition of "FAINT or lower," then create two `character_vitals` objects, one with "breathing = FAINT" and the other (with the same character_id) of "breathing = NONE".
- This should work for persistent characters as well.  Try setting `persistent_characters` to True in the scene and create a condition on a character from the previous scene.
- You can see documentation (for the TA1s) of transition conditions [here](https://nextcentury.atlassian.net/wiki/spaces/ITMC/pages/3041951763/Scenario+YAML+Documentation#Condition-level-elements).  Please make sure it makes sense and looks good.
- To test, you can use [metrics-adept-eval-test.yaml.txt](https://github.com/NextCenturyCorporation/itm-evaluation-server/files/15068747/metrics-adept-eval-test.yaml.txt) (remove the .txt extension) and put it in `swagger_server/itm/data/metrics/scenarios`.    Tweak the scenario as you like to test different paths, running it against the development client:
  - `python itm_minimal_runner.py --adm_name foobar --session adept --scenario adept-test-1`


kaitlyn-sharo (approved):

swagger/swagger.yaml
```diff
@@ -1028,13 +1028,13 @@ components:
             - adept-september-demo-probe-1-choice2
         character_vitals:
           type: array
-          description: True if the specified list of vitals values have been met for the specified character_id
+          description: True if the any of the specified collection of vital values have been met for the specified character_id
```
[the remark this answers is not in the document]
nextcen-dgemoets: Fixing a typo is not nit-picky, thank you!  And in fact, it turns out I had forgotten to re-generate models, so I went ahead and did that too.  Wouldn't have caught that without your nit-picky comment.  :)

swagger_server/itm/itm_scene.py
```diff
@@ -157,11 +162,75 @@ def _actions_condition_met(self, actions_condition) -> bool:
                 return True
         return False
 
+    # True if any of the specified supplies is <= the specified quantity
+    def _supply_condition_met(self, supply_conditions :List[Supplies], session_state :State) -> bool:
```
dereknop, line 166: Everything worked great for me. I just learned how to use the any() function for conditions over the weekend though so here is a suggestion that you can take or leave if you're interested and don't think it would hurt maintainability. 

```
def _supply_condition_met(self, supply_conditions: List[Supplies], session_state: State) -> bool:
        if not supply_conditions:
            return True

        supply_conditions_met = any(
            state_supply.quantity <= supply_condition.quantity 
            for supply_condition in supply_conditions 
            for state_supply in session_state.supplies 
            if supply_condition.type == state_supply.type
        )
       if supply_conditions_met:
          print(f'_supply_condition_met: a supply condition is met; returning true')
       else:
          print(f'_supply_condition_met: did not find a supply meeting criteria; returning false')

       return supply_conditions_met
```

dereknop (approved):

swagger_server/itm/itm_scene.py
nextcen-dgemoets: Love it.  I bet I have other places I can use something like that.

PR #61, opened 2024-04-29
ITM-357: Convert prints to proper logging

nextcen-dgemoets:
Use standard Python logging instead of prints for logging.  Better late than never.
Re-evaluated in some cases what log level something should be, but open to suggestions.


dereknop (approved):
Looks good to me, logs show up fine and is still easy to read

PR #62, opened 2024-05-02
ITM-383: Return history to ADM after each training scenario

nextcen-dgemoets:
Return the full stringified JSON scenario history in the `unstructured` of the final returned `State` when a training scenario ends.  There is an internal toggle for it, which would be false by default, but enabled during training sessions.  It could also be enabled locally (by developers) for debugging non-training sessions.

I also took this opportunity to update/fix the "proper treatments".

To test, you can use the `development` client's `itm_minimal_runner.py` but note you'll need to remove line ~214 where we only print out the final state for non-training scenarios.

### DIFF
--- a/swagger/swagger.yaml
+++ b/swagger/swagger.yaml
@@ -1006,7 +1006,7 @@ components:
           description: True if the scenario elapsed time (in seconds) is greater than the specified value
           minimum: 5
         actions:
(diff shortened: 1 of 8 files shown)


kaitlyn-sharo (approved):
Code and results look great to me

PR #63, opened 2024-05-07
ITM-377: Change scene int index to string ID

nextcen-dgemoets:
- Changed integer `Scene.index` to string `Scene.id`.
- Changed Boolean `Scene.final_scene` to string `Scene.next_scene`.
- Added string `ActionMapping.next_scene`.
- Added optional `Scenario.first_scene`, which defaults to the first scene in the file.

Scenes are now string IDs instead of indexes, which had to be sequentially listed in the YAML as index 0 to n (which was inflexible and error-prone).

`Scene.next_scene` indicates the default next scene for action mappings.  If `Scene.next_scene` is omitted, then by default, this is the last scene in the scenario.  However, to support simple indexed scenes with no branching paths (and improve backward-compatibility), if a scene ID is a non-negative integer `n`, then by default, the next scene `n+1`.

Either way, the scene default can be overridden for individual action mappings via `ActionMapping.next_scene`.  This also means it's possible to end a scene by default, but have certain actions continue the scene (e.g., if they evacuate then the scenario ends, but if they continue treatment then it doesn't).

To test, use the client `development` branch to run `soartech` and `adept` sessions.  You can specify `--count n` to run a total of `n` scenarios, so `--count 500` will test more branches/cases than without `--count`.  Also, look at the scenario YAML files and ensure that no scenario behavior changed.

Also please take a look at the [Scenario YAML documentation](https://nextcentury.atlassian.net/wiki/spaces/ITMC/pages/3041951763/Scenario+YAML+Documentation) and ensure it reflects these changes.  If you prefer, you can [view the diffs](https://nextcentury.atlassian.net/wiki/pages/diffpagesbyversion.action?pageId=3041951763&selectedPageVersions=26&selectedPageVersions=23).

### DIFF
--- a/swagger/swagger.yaml
+++ b/swagger/swagger.yaml
@@ -374,6 +374,9 @@ components:
           type: string
           description: human-readable scenario name, not necessarily unique
           example: IED Explosion
+        first_scene:
+          type: string
+          description: indicates the first/opening scene ID in the scenario
         session_complete:
           type: boolean
           description: "set to true if the session is complete; that is, there are\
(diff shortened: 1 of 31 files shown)


dereknop (approved):
ran with --count 500 for adept and soartech and didn't run into any errors. I did notice random pauses when it was running, probably something on my end but figured I'd mention it. Code looks good 👍

kaitlyn-sharo (approved):
I know we might be tabling this for now, but everything does seem to look good and work well

nextcen-dgemoets:
This enhancement was [met with approval](https://inthemoment-group.slack.com/archives/C05AQMCA5PB/p1715360763600539) from the TA1s, so it looks like we're doing it.

PR #65, opened 2024-05-15
ITM:401: Indicate to ADMs that a blanket has been applied

nextcen-dgemoets:
Added `Character.hasBlanket` so that ADMs can tell if a blanket has been applied.  Normally treatments are conveyed via the `Injury.status` set to `treated`, but Blankets don't treat an injury (but do need to be reflected in the character).
Also added the new supplies to freeform scenarios.

Also, ninja edit to add treatment to Forehead Scape / Abrasion now that it's implemented in the sim, see ITM-406.

To test, use the Human input simulator in the [ITM-401](https://github.com/NextCenturyCorporation/itm-evaluation-client/tree/ITM-401) branch of the `itm-evaluation-client`:
- `python itm_human_input.py --session adept --scenario freeform-desert`

You don't need to run a TA1 server.
The Human input simulator is a little wonky if you've never used it before.  But basically:
1.  Type `s` then `c` to start the session and scenario.
2.  Type `v`, then `t` to take an action.  Then select one of the "treat" actions by number (e.g., `7, treat-kimberly`).
3.  Select any number for the injury location.
4.  Select the `Blanket` treatment by number.
5.  Type a justification or hit enter.
6.  Scan the displayed `State` and ensure that a `Blanket` supply was used, that Kimberly's `hasBlanket` is true (and the others' is still false), and that the elapsed time is 10.
7. Repeat steps 2-5 (trying to apply another blanket to the same patient).
8. Scan the displayed `State` and ensure that a `Blanket` supply was NOT used (it should be 998), that Kimberly's `hasBlanket` is stiill true (and the others' is still false), and that the elapsed time is 20 (10 for the blanket and 10 for a re-treatment).
9. Type `q` to quit.

[ITM-401]: https://nextcentury.atlassian.net/browse/ITM-401?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/swagger_server/itm/itm_action_handler.py
+++ b/swagger_server/itm/itm_action_handler.py
@@ -230,13 +230,21 @@ class ITMActionHandler:
         # If the treatment treats the injury at the specified location, then change its status to treated.
         supply_used = action.parameters.get('treatment', None)
         attempted_retreatment = False
-        for injury in character.injuries:
-            if injury.location == action.parameters.get('location', None):
-                if injury.status != InjuryStatusEnum.TREATED: # Can't attempt to treat a treated injury
-                    if self._proper_treatment(supply_used, injury.name, injury.location):
-                        injury.status = InjuryStatusEnum.TREATED
-                else:
-                    attempted_retreatment = True
+        doesnt_treat_injuries = [SupplyTypeEnum.BLANKET, SupplyTypeEnum.BLOOD, SupplyTypeEnum.EPI_PEN, \
+                                 SupplyTypeEnum.IV_BAG, SupplyTypeEnum.PAIN_MEDICATIONS, SupplyTypeEnum.PULSE_OXIMETER]
+        if supply_used not in doesnt_treat_injuries:
+            for injury in character.injuries:
+                if injury.location == action.parameters.get('location', None):
+                    if injury.status != InjuryStatusEnum.TREATED: # Can't attempt to treat a treated injury
+                        if self._proper_treatment(supply_used, injury.name, injury.location):
+                            injury.status = InjuryStatusEnum.TREATED
+                    else:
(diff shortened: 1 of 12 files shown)


dereknop (approved):
Following the steps, everything checked out as you described (my times elapsed were 10 seconds longer because I did a check vitals action first). It would be handy if there was some text in bold after the state object is returned that highlights the change to the character's state that you just treated so we don't have to scan the whole object.

swagger_server/itm/itm_action_handler.py
```diff
@@ -230,13 +234,21 @@ def apply_treatment(self, action: Action, character: Character):
...
         supply_used = action.parameters.get('treatment', None)
         attempted_retreatment = False
-        for injury in character.injuries:
-            if injury.location == action.parameters.get('location', None):
-                if injury.status != InjuryStatusEnum.TREATED: # Can't attempt to treat a treated injury
-                    if self._proper_treatment(supply_used, injury.name, injury.location):
-                        injury.status = InjuryStatusEnum.TREATED
-                else:
-                    attempted_retreatment = True
+        doesnt_treat_injuries = [SupplyTypeEnum.BLANKET, SupplyTypeEnum.BLOOD, SupplyTypeEnum.EPI_PEN, \
```
brianpippin, line 237: Curious about the inclusion of some supplies here, is it just because we haven't defined the specific injuries yet?  I would assume blood would treat an injury or condition of "loss of blood" and that an epi pen would treat an allergic reaction.  Is this just a temporary hold spot for those until the rest is implemented?

brianpippin (approved):

PR #67, opened 2024-05-20
ITM-386: Develop a way for ADMs to express intent to perform an action

nextcen-dgemoets:
Changes:
- Added `/ta2/intend_action` call to API
- Reject (via 400 error code) when there's a mismatch of action type (e.g., using `/ta2/takeAction` for an intent action and vice versa)
- When all actions are intent actions, don't add unmapped action types.
- Added test scenario ( **to be removed prior to merge** )

Check out the [corresponding client PR](https://github.com/NextCenturyCorporation/itm-evaluation-client/pull/33).

To test:
- Use the [ITM-386 client branch](https://github.com/NextCenturyCorporation/itm-evaluation-client/tree/ITM-386)'s Human input simulator:
  - `python itm_human_input.py --session adept --scenario adept-test-1`
- Type commands `s`, `c`, `v`, then `i` to *intend* an action
- Type any command by number, then optionally a justification.
  - Ensure you get a `400` error because you cannot **take** actions via `intend_action`.
- Type `v`, then `t` to *take* an action, then any command by number and an optional justification.
  - Ensure you don't get an error.
- Type `v` and ensure that the action has `intent_action` set to `True`.
- Type `t` to *take* an action, then any command by number and an optional justification.
  - Ensure you get a `400` error because you cannot **intend** actions via `take_action`.
- Type `v`, then `i` to *intend* an action, then any command by number and an optional justification.
  - Ensure you don't get an error.
- If you like, you can complete the scenario by taking any action, then starting the next scenario (`c`) whereupon the session should end.

Feel free to test `soartech`, `adept`, and/or `eval` sessions to make sure nothing broke.

### DIFF
--- a/swagger/swagger.yaml
+++ b/swagger/swagger.yaml
@@ -348,7 +348,48 @@ paths:
               schema:
                 $ref: "#/components/schemas/State"
         "400":
-          description: Invalid action or Session ID
+          description: Invalid action or Session ID, or specified an intent action
+        "500":
+          description: An exception occurred on the server; see returned error string.
+          content:
+            text/plain:
+              schema:
+                type: string
+                x-content-type: text/plain
+      x-openapi-router-controller: swagger_server.controllers.itm_ta2_eval_controller
+  /ta2/intendAction:
+    post:
+      tags:
+        - itm-ta2-eval
+      summary: Express intent to take an action within a scenario
+      description: Express intent to take the specified Action within a scenario
+      operationId: intend_action
+      parameters:
+        - name: session_id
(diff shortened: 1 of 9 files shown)


dereknop (approved):
Properly blocked me from taking an action when `intent_action` was `true` and vice versa. I also tested sessions (including eval) with the minimal runner and didn't experience any unintended side effects.

kaitlyn-sharo (approved):
Everything worked as described. Waiting on explanation (via teams) about why this is how it works, but approved since it works as expected

PR #69, opened 2024-05-24
Itm 376

dereknop:
https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?assignee=60ba425da547eb00686ee0ce&selectedIssue=ITM-376

This ticket builds off the changes of persist_characters. If `persist_characters` is `true`, you can enter an array of character id's as `removed_characters` to be removed. For example, if scene one had characters 'Bob', 'Sally', and 'Joe' and we are transitioning to scene two with `persist_characters` marked as `true`, the default behavior would be for all three of those characters to exist in the scene without needing to be explicitly defined in the YAML. However, if `persist_characters` is true AND `removed_characters` is `['Sally']`, then the two characters persisting into the next scene should be Joe and Bob

Test using the [corresponding branch](https://github.com/NextCenturyCorporation/itm-evaluation-client/pull/35) of itm-evaluation-client.

- Run `python itm_human_input.py --session adept --scenario test-1`
- Enter `s`, then `c`
- `v` for available actions
- `t` to take an action to trigger scene change
- Confirm you can see in the server logs that before the scene change we have Mike as a character but after the scene change Mike does not persist BUT Civilian_01 does (I will remove these logs prior to merge, it just makes testing easier)

Feel free to mess around with the yaml file (swagger_server/itm/data/metrics/scenarios/metrics-adept-eval-test.yaml) to test different removed_characters.

Note: I specifically placed an action 'Check Mike's vitals' in the scene where he should be removed. Please confirm that this action does not appear as an option.


kaitlyn-sharo (approved):
Works with your given scenario and seems to ignore removed_characters when persist_characters is false. I tried to break it and couldn't!

swagger_server/itm/itm_scenario.py
```diff
@@ -188,6 +201,16 @@ def change_scene(self, next_scene_id):
             else:
                 # No characters were specified in the scene, so inherit characters from previous scene.
                 target_state.characters = previous_scene_characters
+            # if removed_characters found in scene, remove those characters from the
+            if hasattr(self.isd.current_scene, 'removed_characters'):
```
nextcen-dgemoets, line 205: Won't this always be true because you initialized this in `itm_scenario_reader` to `[]` if it's missing?  In theory, you should be able to check its length instead, but defensive programming suggests you should check for `None` also, so:
```
if getattr(self.isd.current_scene, 'removed_characters', None) and len(self.isd.current_scene.removed_characters) > 0:
```

nextcen-dgemoets (changes requested):
Removing the characters works, but the filter to make sure we're not proposing an action with an invalid character is over-filtering.

Also, I'm suggesting logging errors for YAML configuration errors (if they get through the validator) because they need to be spotted and reported to TA1.

nextcen-dgemoets (approved):
Changes look good-- thanks.  I suggested that the one message include why the action is being filtered out.

Also, don't forget to remove the test logging.

PR #70, opened 2024-05-29
ITM-385: Add test session and YAML examples

nextcen-dgemoets:
- Added a testing framework / session type for testing PRs and a precursor to regression testing (incompatible with training mode);
- Updated sample YAML file from early in Metrics eval to current development;
- Wrote some YAML examples that use the various features of the scenario YAML file that we can run routinely after every ticket / new feature;
- Made `CHECK_BLOOD_OXYGEN` and `CHECK_ALL_VITALS` fail if there is no Pulse Oximeter in the supplies;
  - Also don't add them as actions if no pulse ox is available.
- Fixed regression introduced in ITM-386.

See corresponding [client PR](https://github.com/NextCenturyCorporation/itm-evaluation-client/pull/36) and [branch](https://github.com/NextCenturyCorporation/itm-evaluation-client/tree/ITM-385).

To test, run test sessions (maybe with a very large `--count` and make sure there are no crashes.  You can also try to understand what's going on in each of the tests and see if it's achieving its goal.

Also feel free to suggest other tests.  I figure we'll add more as we go, but feel free to suggest.  I would like to add one that tests all of the "transition conditions".

### DIFF
--- a/swagger/swagger.yaml
+++ b/swagger/swagger.yaml
@@ -40,7 +40,7 @@ paths:
             type: string
         - name: session_type
           in: query
-          description: "the type of session to start (`eval` or a TA1 name)"
+          description: "the type of session to start (`eval`, `test`, or a TA1 name)"
           required: true
           style: form
           explode: true
--- a/swagger_server/controllers/itm_ta2_eval_controller.py
+++ b/swagger_server/controllers/itm_ta2_eval_controller.py
@@ -115,7 +115,7 @@ def start_session(adm_name, session_type, adm_profile=None, kdma_training=False,
 
     :param adm_name: A self-assigned ADM name.
     :type adm_name: str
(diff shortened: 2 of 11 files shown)


dereknop:
Initial testing with `python3 itm_minimal_runner.py --name foobar --session test --count 2000` appeared like everything was working correctly (no crashes). I still need to dig a bit deeper to understand all of the tests though

dereknop (approved):
Tried to break but didn't run into any issues 👍. 
Side Note: replacing these return statements `return self.times_dict["SEARCH"]` with `return self.times_dict[ActionTypeEnum.SEARCH]` was a nice touch

kaitlyn-sharo (approved):
Everything looks good and works well. For anyone else testing, 100,000 may be a bit too high of a count to test with.

PR #72, opened 2024-06-11
ITM-422: Support "unseen" characters and ability to search/find them

nextcen-dgemoets:
Work performed:
- Add `unseen` property to the `Character` object;
- Add `MOVE_TO` action and implement `SEARCH`;
- Develop unseen characters test YAML; and
- Fixed bug with SITREP character validation

Use with client branch [ITM-422](https://github.com/NextCenturyCorporation/itm-evaluation-client/tree/ITM-422) and its corresponding [PR](https://github.com/NextCenturyCorporation/itm-evaluation-client/pull/38).

To test, run the test session with maybe a `count` of 100 or so.
Then use the human adm simulator and try different paths and see that you get the expected results.  But note the actions that advance the scenes:  `move`, `move`, `search`, `move`, and `end_scene`.

[ITM-442]: https://nextcentury.atlassian.net/browse/ITM-442?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ITM-422]: https://nextcentury.atlassian.net/browse/ITM-422?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


swagger_server/itm/itm_action_handler.py
```diff
@@ -391,8 +404,11 @@ def tag_character(self, character: Character, tag: str):
 
     def search(self):
         """
-        Search for more characters in the scene.
+        Search for more characters.
         """
+        # After a search, the ADM is at a new location (which may or may not have patients), so all previous characters become unseen.
+        for character in self.session.state.characters:
```
dereknop, line 410: You mention that there may or may not be patients in the new location. I think I am just missing where those new potential characters are added to the state because here we are setting all of the characters to `unseen`. Because if there are patients at this new location they wouldn't be marked as unseen, right?

swagger_server/itm/itm_action_handler.py
```diff
@@ -365,6 +365,19 @@ def direct_mobile_characters(self):
         return self.times_dict[ActionTypeEnum.DIRECT_MOBILE_CHARACTERS]
 
 
+    def move_to(self, target_character: Character):
```
dereknop, line 368: we take in a `target_charcter` here but don't use it in the function.

swagger_server/itm/itm_action_handler.py
```diff
@@ -365,6 +365,19 @@ def direct_mobile_characters(self):
...
+        """
+        Move to the location of the specified character, toggling whether all characters are seen or not.
+        NOTE: This only works when there are only two locations, which is the stated requirement.
```
dereknop, line 371: This comment made me think that maybe characters should have a `location` field. In case there are ever more than two locations in the future, we could modify this for loop to set all of the `unseen` values to `true` for characters without the same location as the `target_character`

dereknop (commented):
I didn't encounter any issues while testing but left a few comments that occurred to me.

swagger_server/itm/itm_action_handler.py
nextcen-dgemoets: Yep, the onus is on the scenario designer.  There's no new code to handle that.  Check out the `after_search` scene in the test file and you'll see that `Civilian_02` is added to the state.  And you're right-- the new character should *not* be marked unseen.

swagger_server/itm/itm_action_handler.py
nextcen-dgemoets: Yep, that is quirky.  The API takes in a target character, but with the current two-room-only implementation, it's not actually used in the code.  I can remove it from the `move_to` call if you like, but it just seemed odd to have a `move_to` call with no "direct object," and if we ever implement multiple locations, we'll need to know which character the ADM is moving to.

Or, I can add a comment explaining the above.

swagger_server/itm/itm_action_handler.py
nextcen-dgemoets: If we wanted/needed to support more than two locations, then I'm pretty sure you're right-- we'd need a `Character.location` field.  Brian and I are trying to steer clear from >2 locations because we don't really want ADMs to be bouncing all over the scene, having to keep track of where the patients are, spending "time" moving, etc.  But if it becomes a requirement, then we'll implement it and likely add a `location` field, as you suggest.

swagger_server/itm/itm_action_handler.py
dereknop: Yea, I suppose we leave it because it isn't hurting anything and like you mentioned we may use it later on. Maybe we add a comment here to remind us later.

swagger_server/itm/itm_action_handler.py
dereknop: Makes sense

swagger_server/itm/itm_action_handler.py
nextcen-dgemoets: Added comment.  Agree that it looks odd to have an unused parameter like that.

kaitlyn-sharo (approved):
Works great

PR #75, opened 2024-06-24
ITM-461

dereknop:
[ITM-461](https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?assignee=60ba425da547eb00686ee0ce&selectedIssue=ITM-461)

Before the warning log would never appear because the removed characters were filtered out before checking if there was a matching character id in the `target_state` (so the list was always empty). I am checking for matching character id's in the `target_state` and `removed_characters` before filtering them out to trigger a warning log. The character will still be removed from the scene. To test, I used the `removed_characters.yaml` file. You'll see that when we transition into the second scene (scene.id: 1), Mike is listed as a character in the target state as well as in `removed_characters`. The sever should warn that this is a misconfiguration and then still remove him from the scene. I have added an additional log for testing that will list the current characters when the scene changes. For this test file the characters after the scene change should be: `Civilian_01` and `Captain_01`

There are no changes to the client, so testing with the client's dev branch is fine.

`python itm_minimal_runner.py --name foobar --session test --scenario removed_characters`

Ensure that you can see the proper logs, `itm_human_input.py` could be used for more in-depth testing.

[ITM-461]: https://nextcentury.atlassian.net/browse/ITM-461?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/swagger_server/itm/itm_scenario.py
+++ b/swagger_server/itm/itm_scenario.py
@@ -100,8 +100,6 @@ class ITMScenario:
         # If in training mode, record probe response in meta info
         if self.session.kdma_training:
             self.session.state.meta_info.probe_response = response
-            print("Probe response in meta info:")
-            print(self.session.state.meta_info.probe_response)
 
         self.session.history.add_history(
             "Respond to TA1 Probe",
@@ -218,15 +216,6 @@ class ITMScenario:
 
             # 1a. Remove `removed_characters`, even if in configured scene state.
             if getattr(self.isd.current_scene, 'removed_characters', None) and len(self.isd.current_scene.removed_characters) > 0:
-                current_state.characters = [
-                    character for character in current_state.characters
-                    if character.id not in self.isd.current_scene.removed_characters
-                ]
-                # if the target_state includes a character that is listed in removed_characters, that is a yaml misconfiguration
-                target_state.characters = [
-                    character for character in target_state.characters
-                    if character.id not in self.isd.current_scene.removed_characters
-                ]
                 filtered_out_characters = [
                     character for character in target_state.characters
                     if character.id in self.isd.current_scene.removed_characters
@@ -234,6 +223,18 @@ class ITMScenario:
 
                 if len(filtered_out_characters) > 0:
                     logging.warning("\033[92mScene configuration issue: target state includes character that was removed\033[00m")
+                
+                current_state.characters = [
+                    character for character in current_state.characters
+                    if character.id not in self.isd.current_scene.removed_characters
+                ]
+                logging.warning("Current characters after scene change")
+                logging.warning(current_state.characters)
+                # if the target_state includes a character that is listed in removed_characters, that is a yaml misconfiguration
+                target_state.characters = [
+                    character for character in target_state.characters
+                    if character.id not in self.isd.current_scene.removed_characters
+                ]
         else:
             current_state.characters = deepcopy(target_state.characters)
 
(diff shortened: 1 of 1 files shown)


dereknop:
> Derek-- looks good. Elegant solution that I don't know why I didn't think of earlier.
> 
> My only question is if this logging (now that it appears) is sufficient, and maybe we don't need the spammier logging in `get_available_actions()`. Is there any way we can get one logging without the other? If so, maybe just keeping the one in `change_scene()` is sufficient.
> 
> Thoughts?

We can use the one warning log in `change_scene` instead of both. However, the spammier one you are referring to is triggered if an entry in the `action_mapping` has a `character_id` that matches a removed character. So they don't _exactly_ indicate the same thing. With that being said, I don't want the warning logs to seem spammy and people to not pay attention to them consequently so I can remove that warning if you think it's best.

nextcen-dgemoets:
> We can use the one warning log in change_scene instead of both. However, the spammier one you are referring to is triggered if an entry in the action_mapping has a character_id that matches a removed character. So they don't exactly indicate the same thing. With that being said, I don't want the warning logs to seem spammy and people to not pay attention to them consequently so I can remove that warning if you think it's best.

Are you saying it's possible to get one without the other?  If that's the case, I think we can keep the spammier message, but make it a bit less spammy.  How about:
- "Scene configuration issue: ignoring action <action_id> with invalid character <character_id>"

I'm a bit on the fence, so you can convince me to remove it entirely.  If it's impossible to get one without the other, then remove the spammier one.

dereknop:
> > We can use the one warning log in change_scene instead of both. However, the spammier one you are referring to is triggered if an entry in the action_mapping has a character_id that matches a removed character. So they don't exactly indicate the same thing. With that being said, I don't want the warning logs to seem spammy and people to not pay attention to them consequently so I can remove that warning if you think it's best.
> 
> Are you saying it's possible to get one without the other? If that's the case, I think we can keep the spammier message, but make it a bit less spammy. How about:
> 
> * "Scene configuration issue: ignoring action <action_id> with invalid character <character_id>"
> 
> I'm a bit on the fence, so you can convince me to remove it entirely. If it's impossible to get one without the other, then remove the spammier one.

I thought it would be possible to get one without the other (if a removed character wasn't in the scene state but was in the action mapping) but upon testing, I always get both. So I will go ahead and remove the more annoying warning log. Thanks!

swagger_server/itm/itm_scenario.py
```diff
@@ -70,17 +70,10 @@ def get_available_actions(self) -> List[Action]:
 
         # safe guarding that an action with character id of a removed character doesn't slip through the cracks
         filtered_actions = []
-        filtered_out_actions = []
 
         for action in actions:
             if not getattr(action, 'character_id', None) or action.character_id in current_character_ids:
                 filtered_actions.append(action)
-            else:
```
nextcen-dgemoets, line 78: Could you add something like:
```
else:
    pass # This is an error condition that will be flagged when changing scenes
```

nextcen-dgemoets (changes requested):
One small suggestion, then remove the warning logging in `change_scene()`.  Then I'll switch to approved.  Thanks!

nextcen-dgemoets (approved):

PR #76, opened 2024-06-27
ITM-471: Added Events from scenario to ADM.

nextcen-dgemoets:
Use with [client PR 41](https://github.com/NextCenturyCorporation/itm-evaluation-client/pull/41).

To test, use an ADM simulator to ensure that two events appear in the state in the "evac_decision" scene, and one (new) event appears in the "cleanup" scene.

Also contains a small fix to ITM-461.


dereknop (approved):
Works as described 👍

swagger/swagger.yaml
```diff
@@ -739,6 +744,39 @@ components:
           $ref: "#/components/schemas/ThreatTypeEnum"
         severity:
           $ref: "#/components/schemas/ThreatSeverityEnum"
+    Event:
+      required:
+        - unstructured
+        - type
+      type: object
+      description: an unit of structured communication from scenario to ADM
```
brianpippin, line 752: This is so nitpicky, but it should be "a unit" instead of an (Sorry don't kill me):

For words where the “u” sound is pronounced like “you,” such as user, usual, or utilized, use “a.”
For words where the “u” sound is pronounced like “uh,” such as unusual, understanding, or utter, use “an.”

nextcen-dgemoets: Not too nitpicky!  I agree and I'll fix it!  (Thought I might fix it in ITM-472 depending how many other issues there are.)

brianpippin (approved):
Looks good, didn't see anything.

swagger/swagger.yaml
nextcen-dgemoets: This is addressed in ITM-477's [PR](https://github.com/NextCenturyCorporation/itm-evaluation-server/pull/77).

PR #83, opened 2024-07-09
ITM-490: Transition to Dryrun evaluation

nextcen-dgemoets:
Added infrastructure for dryrun evaluation, especially available TA1 scenarios.

It's possible we'll still make the following changes prior to merging this PR:
- ~~Removing KDMA values from evaluation scenarios;~~
- ~~Removing evaluation scenarios;~~
- ~~Removing training scenarios;~~
- ~~Updating scenarios or replacing provisional local alignment targets~~; and/or

To test, feel free to scan the YAML for anything that looks odd, but basically if our clients are able to run through the scenarios without errors, then that's good enough for our purposes (to hand them over to TA2).

For testing, first I'll suggest setting the two variables after `if kdma_training` in `start_session` to `False`, so that it doesn't try to connect to a TA1 server.  (Approx lines 428/429)  Then, I'll suggest the following commands to exercise the scenarios:
- `python itm_minimal_runner.py --session adept --name foobar  --train --count 100`
- `python itm_minimal_runner.py --session soartech --name foobar  --train --count 100`
- `python itm_minimal_runner.py --session soartech --name foobar  --count 100`

There's no adept without training because there are no eval scenarios available from ADEPT at the moment.


kaitlyn-sharo (approved):

dereknop (approved):
LGTM. No errors running through the sessions

Ayal-Ciobotaru (approved):

PR #86, opened 2024-07-11
ITM-510

Ayal-Ciobotaru:
Added 'justification' as a requirement for `action` and added `medical_policies` controlled vocab


Ayal-Ciobotaru:
Adjusting based on discussion with Darren

swagger/swagger.yaml
```diff
@@ -1311,6 +1313,14 @@ components:
         - some
         - extensive
         - crowd
+    MedicalPoliciesEnum:
+      type: string
+      description: Directives issued by competent military authority
```
nextcen-dgemoets, line 1318: Could you change "Neutral" to "Treat All Neutrally"?

swagger/swagger.yaml
```diff
@@ -609,8 +609,10 @@ components:
           type: string
           description: The political climate in a mission to inform decision-making
         medical_policies:
-          type: string
-          description: Medical policies in effect in a mission, to inform decision-making
+          type: array
+          description: A list of medical policies
```
nextcen-dgemoets, line 613: Could you add, "; omit this property if no special policy is in place."

swagger/swagger.yaml
```diff
@@ -348,7 +348,7 @@ paths:
               schema:
                 $ref: "#/components/schemas/State"
         "400":
-          description: Invalid action or Session ID, or specified an intent action
+          description: Invalid action or Session ID, empty justification or specified an intent action
```
nextcen-dgemoets, line 351: Although this is true, we're not listing any other reasons the action can be invalid...so probably just leave it how it was before.

nextcen-dgemoets (commented):
Obviously I'll test after the client PR is re-opened, but here are some preliminary comments.

swagger_server/itm/data/dryrun/test/sample.yaml
```diff
@@ -19,7 +19,7 @@ state:
     communication_capability: both # current availability of internal and external communication, controlled vocab includes internal, external, both, neither
     roe: unused # rules of engagement to inform decision-making, but not to restrict decision space
     political_climate: unused # The political climate in a mission to inform decision-making
-    medical_policies: unused # Medical policies in effect in a mission, to inform decision-making
+    medical_policies: Treat All Neutrally # controlled vocab includes Treat All Neutrally, Treat Enemy LLE, Treat Civilian LLE, and Prioritize Mission
```
nextcen-dgemoets, line 22: This string needs to be converted into an array with one string.

swagger_server/itm/data/metrics/test/sample.yaml
```diff
@@ -19,7 +19,7 @@ state:
     communication_capability: both # current availability of internal and external communication, controlled vocab includes internal, external, both, neither
     roe: unused # rules of engagement to inform decision-making, but not to restrict decision space
     political_climate: unused # The political climate in a mission to inform decision-making
-    medical_policies: unused # Medical policies in effect in a mission, to inform decision-making
+    medical_policies: Treat All Neutrally # controlled vocab includes Treat All Neutrally, Treat Enemy LLE, Treat Civilian LLE, and Prioritize Mission
```
nextcen-dgemoets, line 22: This string needs to be converted into an array with one string.

nextcen-dgemoets (changes requested):
Need to fix the sample; otherwise looks good.

nextcen-dgemoets (approved):

dereknop (approved):

PR #88, opened 2024-07-12
ITM-512

Ayal-Ciobotaru:
Created a `config_util` class that handles:

- checking to make sure the config file exists 
- reading in the config file
- verifying the expected keys are in the config file

Added a `.template` version of the config.ini so that each user can create their own config.ini file and not get it overwritten by different commits. In order to get it working, copy over the `config.ini.template` file, renaming it to `config.ini`

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,19 @@ On Windows, the method to activate depends on the shell:
 - PowerShell: `venv\Scripts\Activate.ps1`
 - cmd.exe: `venv\Scripts\activate.bat`
 
+
+## Configuration
+
+Rename `config.ini.template` file to `config.ini`. 
+
+The following properties can be configured:
+- `EVALUATION_TYPE` 
+    - default is `dryrun`
+- `SCENARIO_DIRECTORY`
+    - default is `swagger_server/itm/data/%(EVALUATION_TYPE)s/scenarios/`
+
+*NOTE:* the trailing **`s`** in `.../data/%(EVALUATION_TYPE)s/...` is needed for string interpolation to work properly.
+
 ## Usage
 To run the server, please execute the following from the root directory:
 
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Rename `config.ini.template` file to `config.ini`.
 
 The following properties can be configured:
 - `EVALUATION_TYPE` 
-    - default is `dryrun`
+    - default is `dryrun` but `metrics` is also supported
 - `SCENARIO_DIRECTORY`
     - default is `swagger_server/itm/data/%(EVALUATION_TYPE)s/scenarios/`
 
--- a/swagger_server/config.ini.template
+++ b/swagger_server/config.ini.template
@@ -1,0 +1,7 @@
+; Copy over this template file and rename it to config.ini
+; Replace the default configurations below with the desired values
+
+[DEFAULT]
+
+EVALUATION_TYPE=dryrun
+SCENARIO_DIRECTORY=swagger_server/itm/data/%(EVALUATION_TYPE)s/
--- a/swagger_server/config.ini.template
+++ b/swagger_server/config.ini.template
@@ -4,4 +4,4 @@
 [DEFAULT]
 
 EVALUATION_TYPE=dryrun
-SCENARIO_DIRECTORY=swagger_server/itm/data/%(EVALUATION_TYPE)s/
+SCENARIO_DIRECTORY=swagger_server/itm/data/%(EVALUATION_TYPE)s/scenarios/
--- a/swagger_server/config_util.py
+++ b/swagger_server/config_util.py
@@ -1,6 +1,6 @@
 import os
 from os.path import exists
-from configparser import ConfigParser, NoOptionError
+from configparser import ConfigParser
 from dataclasses import dataclass
 from typing import Optional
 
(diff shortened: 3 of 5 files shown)


swagger_server/config.ini.template
```diff
@@ -0,0 +1,7 @@
+; Copy over this template file and rename it to config.ini
+; Replace the default configurations below with the desired values
+
+[DEFAULT]
+
+EVALUATION_TYPE=dryrun
+SCENARIO_DIRECTORY=swagger_server/itm/data/%(EVALUATION_TYPE)s/
```
nextcen-dgemoets, line 7: Haven't actually tried this out yet, but is that trailing `s` supposed to be there?  The folders are currently `metrics` and `dryrun`, but it looks like you're unconditionally adding a trailing `s` to the path.  As written, wouldn't this require the path be `swagger_server/itm/data/dryruns/`?

Ayal-Ciobotaru: the trailing `s` is supposed to be there, it confused me as well when I read the documentation. That is how the `.ini` files do value interpolation. https://docs.python.org/3/library/configparser.html#interpolation-of-values

swagger_server/itm/itm_session.py
```diff
@@ -447,7 +452,7 @@ def start_session(self, adm_name: str, session_type: str, adm_profile: str, kdma
                 self._end_session() # Exception here ends the session
                 return 'Exception communicating with TA1; is the TA1 server running?  Ending session.', 503
 
-        path = f"swagger_server/itm/data/{ITMSession.EVALUATION_TYPE}/" + ("test/" if self.session_type == 'test' else 'scenarios/')
+        path = f"{ITMSession.SCENARIO_DIRECTORY}" + ("test/" if self.session_type == 'test' else 'scenarios/')
```
nextcen-dgemoets, line 455: A few things:
- In the ticket, we talked about Test scenarios always being in `itm/data/{evaluation}/test`.  Those are used during `test` sessions.  So use `EVALUATION_TYPE` but not `SCENARIO_DIRECTORY` for `test` sessions.
- It seems we could add a `/` after `{ITMSession.SCENARIO_DIRECTORY}` so that their paths work whether or not they specify a trailing slash.
- Maybe we shouldn't require their paths have a `scenarios` subdirectory.  In other words, if they specify `foobar`, we should look for scenarios in `./foobar/` instead of `./foobar/scenarios/`.  The default would need to change to `swagger_server/itm/data/%(EVALUATION_TYPE)s/scenarios/`.

nextcen-dgemoets (changes requested):
First of all, **I love this!!!!!*   And so will TA2.

A few comments:
- Need to update the README to document this configuration.  Just come up with something in a new section, and I may tweak wording because I can't help myself.
- Could you add the scenario path to the logging at the end of `start_session`?  This way, it will say something like "Loaded 8 total scenarios from swagger_server/itm/data/dryrun/scenarios."  It might help troubleshoot if you're pointing someplace you're not expecting to load your scenarios.
- I like this so much, I might want to add more to config, like whether or not we are trying to connect to TA1 servers.  But I'll write a new ticket for that and maybe address it later.
- I have some additional suggestions inlined, especially about how the paths are formed.

Thank you!

README.md
```diff
@@ -25,6 +25,19 @@ On Windows, the method to activate depends on the shell:
...
 - cmd.exe: `venv\Scripts\activate.bat`
 
+
+## Configuration
+
+Rename `config.ini.template` file to `config.ini`. 
+
+The following properties can be configured:
+- `EVALUATION_TYPE` 
+    - default is `dryrun`
```
nextcen-dgemoets, line 35: how about "but `metrics` is also supported"

swagger_server/config_util.py
```diff
@@ -0,0 +1,36 @@
+import os
+from os.path import exists
+from configparser import ConfigParser, NoOptionError
```
nextcen-dgemoets, line 3: `NoOptionError` isn't used.

swagger_server/itm/itm_session.py
```diff
@@ -486,7 +491,7 @@ def start_session(self, adm_name: str, session_type: str, adm_profile: str, kdma
                 random_index = random.randint(0, num_read_scenarios - 1)
                 self.itm_scenarios.append(deepcopy(self.itm_scenarios[random_index]))
 
-        logging.info('Loaded %d total scenarios.', len(self.itm_scenarios))
+        logging.info('Loaded %d total scenarios from %s.', len(self.itm_scenarios), ITMSession.SCENARIO_DIRECTORY)
```
nextcen-dgemoets, line 494: This isn't true for `test` sessions anymore.

nextcen-dgemoets (changes requested):
Looks good!  Just a few minor things left.

phile-caci (approved):
LGTM!

Started the Eval Server and Client.
Tested EVALUATION_TYPE=metrics and EVALUATION_TYPE=dryrun with:
`python itm_minimal_runner.py --session test --name foobar  --count 5`
**No issues**

Started Adept server changed EVALUATION_TYPE=metrics
`python itm_minimal_runner.py --session adept --name foobar  --count 5`
**No issues**

Code looks good, nothing stands out.

PR #90, opened 2024-07-16
ITM-523: Added justify message type.  Removed all tbd placeholders.

nextcen-dgemoets:
Use with [client PR](https://github.com/NextCenturyCorporation/itm-evaluation-client/pull/52).

To test, run the sample scenario and ensure that the justify message gets sent without errors.
- `python itm_minimal_runner.py --name foobar --session test --scenario sample-1`


swagger_server/itm/data/dryrun/test/sample.yaml
```diff
@@ -186,13 +186,13 @@ scenes:
     end_scene_allowed: false
     next_scene: evac_decision
     probe_config:
-      - probe_id: adept-september-demo-probe-1
```
Ayal-Ciobotaru, line 189: Not sure if this is necessary, however the other `sample.yaml` located in `itm/data/metrics/test` still has `adept-september-demo-probe-X`

Ayal-Ciobotaru:
I ran `python itm_minimal_runner.py --name foobar --session test --scenario sample-1` with both `metrics` and `dryrun`. No errors came up in the server however the `metrics` type did output with the old probe ids. 

`Responding to probe adept-september-demo-probe-2 from scenario sample-1 with choice s1-p2-choice1.`

swagger_server/itm/data/dryrun/test/sample.yaml
nextcen-dgemoets: Yeah, my thinking (hope?) is that we're not going to continue to update metrics scenarios as the API evolves, including the sample files.  That said, this is a good thing to check with @brianpippin .  Do we need these to stay in a runnable state?

nextcen-dgemoets: I went ahead and updated the metrics samples.  Easy to do, and no harm in it.

dereknop (approved):

Ayal-Ciobotaru (approved):

PR #91, opened 2024-07-17
ITM-478/521/522:  Integrate with TA1 Dry Run servers

nextcen-dgemoets:
- Removed `kdma_values` from `AlignmentResults`, in accordance with program changes. (ITM-478)
- Renamed various IDs to what TA1 is actually using (e.g., alignment targets, kdmas)
- Better handling/logging of TA1 server errors.
- Changed ADM output filenames since we're no longer using high/low
- Bugfixes.

To test, the best case scenario is to run the latest ADEPT and SoarTech servers and run the various session types against them.
If you haven't already, you'll have to copy the TA1 scenarios to your configured location.  I'll need to send them to you under separate cover because they have KDMAs unredacted so cannot be posted publicly.

I would change ITMSession's `self.ta1_integration` to True, then run the various TA1 session types:
- `python itm_minimal_runner.py --name foobar --session adept`
- `python itm_minimal_runner.py --name foobar --session adept --training`
- `python itm_minimal_runner.py --name foobar --session soartech`
- `python itm_minimal_runner.py --name foobar --session soartech --training`

You'll see server errors (especially for SoarTech scenarios) and not all SoarTech scenarios return session alignment.  All of the ADEPT ones should, although they're still updating `DryRunEval-MJ5-eval`.

See [client PR](https://github.com/NextCenturyCorporation/itm-evaluation-client/pull/53)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,7 @@ itm_history_output
 
 # local config files
 config.ini
+
+# local scenarios
+scenarios
+
--- a/README.md
+++ b/README.md
@@ -38,6 +38,17 @@ The following properties can be configured:
 
 *NOTE:* the trailing **`s`** in `.../data/%(EVALUATION_TYPE)s/...` is needed for string interpolation to work properly.
 
+### Scenario filename convention
+Scenario files must be named in the following format to be read by the server at runtime (without punctuation except the indicated hyphens).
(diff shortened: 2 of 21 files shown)


brianpippin (approved):

dereknop (approved):
Looks good to me, only got errors where it was expected

PR #93, opened 2024-07-18
ITM-525: Clean out old models and update references to them.

nextcen-dgemoets:
To test, you can use the itm_human_input and try treating a "neck" location:
- `python itm_human_input.py --session test --scenario sample-1`
- In the first scene, select the `treat-civ` action and select the `neck` location.  You can try the same thing on `main` and you should get a crash.

See [client PR](https://github.com/NextCenturyCorporation/itm-evaluation-client/pull/55).


phile-caci (approved):
LGTM!  tested the given scenario (then played around some more), didn't see any issues.

dereknop (approved):

PR #94, opened 2024-07-19
ITM-526:  Add action pre-validation step 

nextcen-dgemoets:
This happens before action validation, and is to ensure ADM didn't change a pre-configured action in ways it shouldn't.  Right now, the server is very trusting...

To test, what I did was to modify the `itm_minimal_runner.py` client to hack up the action it sends.  After `action = get_next_action()`, add something like:
```
if random.randint(0, 5) == 3:
      action.action_type = "DIRECT_MOBILE_CHARACTERS"
```

Then run, say, an adept session like `python itm_minimal_runner.py --session adept --name foobar`.
Eventually, you should get an error about the `action_type` being wrong.
Try replacing the second line above with other hacks (one at a time) to get different error messages:
- `action.character_id = 'foobar'`
- `action.intent_action = True`
- `action.parameters = {"foobar": "asdf"}`
- `action.parameters = None`

Sometimes you may have to try a few times to get the error message you want.

### DIFF
--- a/swagger_server/itm/itm_session.py
+++ b/swagger_server/itm/itm_session.py
@@ -533,6 +533,53 @@ class ITMSession:
         return self.take_or_intend_action(body, True)
 
 
+    def prevalidate_action(self, adm_action: Action) -> str:
+        """
+        Ensure that an ADM's version of a given action (by action_id) matches the scene's.
+
+        Args:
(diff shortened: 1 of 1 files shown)


Ayal-Ciobotaru (approved):

dereknop (approved):
Works 👍

PR #96, opened 2024-07-24
ITM-478/ITM-521: Integrate with updated SoarTech DryRun server

nextcen-dgemoets:
- Migrate KDMAs to include SoarTech kde's
- Send `user_id` to SoarTech and tweak some logging
- Update API version to 0.3.2 for DRE

See [client PR](https://github.com/NextCenturyCorporation/itm-evaluation-client/pull/57) for testing instructions.


brianpippin (approved):

dereknop (approved):

phile-caci (approved):
Everything looks good on my end also, (after taking away --training)

Adding a comment to the itm_ta1_controller.py (sry, adding it here)

In `def get_alignmetn_data`, is it worth making it so the url setup:

`url = f"http://{host}:{port}/api/v1/{target_id_path}"`

 work with the AWS instances of TA1?:

https://[darpaitm.caci.com/adept/api/v1/alignment_target_ids](https://darpaitm.caci.com/adept/api/v1/alignment_target_ids)
and
https://[darpaitm.caci.com/soartech/api/v1/alignment_targets](https://darpaitm.caci.com/soartech/api/v1/alignment_targets)

That way we can also attach the evaluation server to the TA1 instances on AWS

nextcen-dgemoets:
> In `def get_alignment_data`, is it worth making it so the url setup [will] work with the AWS instances of TA1?
> That way we can also attach the evaluation server to the TA1 instances on AWS

Hmm, so the fact that it hardcodes http is a problem...and possibly the port as well?
Now that we have an actual configuration file for the server, I was thinking of converting the host and port environment variables to config along with the rest of the config...and we can combine them into a single "ADEPT_HOST" which can be a full spec like: "https://darpaitm.caci.com/adept".  Would that address the problem?  Or would it screw up the Docker config?  For that matter, does the Docker config need to copy the config file in the first place?

Is this something we can do later as part of a new ticket?  Or do we need this in order for others to point to the AWS TA3 server.  <or can we hack it up temporarily?>

For that matter, did this work in metrics, or did we have the same problem?

dereknop:
Was able to run eval session with no errors after latest commit

Ayal-Ciobotaru (approved):
lgtm!

nextcen-dgemoets:
> Is this something we can do later as part of a new ticket? Or do we need this in order for others to point to the AWS TA3 server. <or can we hack it up temporarily?>

In discussion with @phile-caci , we determined that what we have now is sufficient to get TA2 started.  The current limitations are that if you use the AWS TA3 server, you are using the AWS TA1 servers.  And you can't run the TA3 server locally and point to the AWS TA1 servers without a code tweak, which we'll describe to TA2s.

This will all be resolved "soon" after we add more config to the .ini file, including combining port and hostname environment variables into hostname config settings, with default value e.g. "https://darpaitm.caci.com/adept".

When we do that, we'll have the `config.ini.template` contain the default values for prod, and have the Dockerfile copy from the template to the .ini file so that deploying to prod is automated.

PR #102, opened 2024-08-20
ITM-535 scenario runner for dry run

phile-caci:
**SRY FORGOT TO MENTION.  Make a new config.ini from config.ini.template**

**LMK if you don't have the updated scenarios, I'll send it in Teams**

1. Updates SoarTech runs to only run the correct alignment target for the corresponding SoarTech scenario.  i.e QOL alignment targets are only for SoarTech QOL scenarios, and same with VOL. This should also half the number of runs from 240 to 120 for SoarTech.
2. Adds evalName/evalNumber to history json
3. Updated filename convention:
  1.  evaluation_type - scenario_id - scene_type - alignment_type-adm-name-timestamp
  2. timestamp was updated from "Aug19" to "20240819"

To Test:
python itm_minimal_runner.py --name foobar --session soartech  

To make sure I didn't break anything run:
python itm_minimal_runner.py --name foobar --session eval  
python itm_minimal_runner.py --name foobar --session adept 

I'll continue testing.


dereknop:
`ERROR 08-20 05:10:34 Contamination in session_alignment! scenario is qol-dre-1-eval but alignment source scenario is qol-dre-1-train.`

I got this when running `python itm_minimal_runner.py --name foobar --session eval`

But nothing crashed. Maybe I have the scenarios under
`swagger_server/itm/data/dryrun/scenarios`

The other two commands ran as I'd expect

phile-caci:
> `ERROR 08-20 05:10:34 Contamination in session_alignment! scenario is qol-dre-1-eval but alignment source scenario is qol-dre-1-train.`
> 
> I got this when running `python itm_minimal_runner.py --name foobar --session eval`
> 
> But nothing crashed. I have the scenarios under `swagger_server/itm/data/dryrun/scenarios`
> 
> The other two commands ran as I'd expect

Thanks @dereknop , I see this on prod also, so I don't think I've broken anything.

dereknop (approved):

brianpippin:
I didn't see any new code for ADEPT.  Do we already have them included in the flag for running the whole evaluation somewhere else?

phile-caci:
> I didn't see any new code for ADEPT. Do we already have them included in the flag for running the whole evaluation somewhere else?

I didn't think we needed a flag for Adept, since they run all alignment targets against all their scenes (which is what it currently does), unlike SoarTech where it's just QOL alignment targets against QOL scenes (and same for VOL).  But let me know if that's not correct.

Ayal-Ciobotaru (approved):

phile-caci:
> > `ERROR 08-20 05:10:34 Contamination in session_alignment! scenario is qol-dre-1-eval but alignment source scenario is qol-dre-1-train.`
> > I got this when running `python itm_minimal_runner.py --name foobar --session eval`
> > But nothing crashed. I have the scenarios under `swagger_server/itm/data/dryrun/scenarios`
> > The other two commands ran as I'd expect
> 
> Thanks @dereknop , I see this on prod also, so I don't think I've broken anything.

@dereknop , was looking into this further. Looks like one of their eval scenarios, dryrun-soartech-eval-qol1.yaml, https://github.com/ITM-Soartech/ta1-server-mvp/blob/dre/ta1_server/data/scenarios/current/dryrun-soartech-eval-qol1.yaml, is calling a training probe.

phile-caci:
> > > I didn't see any new code for ADEPT. Do we already have them included in the flag for running the whole evaluation somewhere else?
> > 
> > 
> > I didn't think we needed a flag for Adept, since they run all alignment targets against all their scenes (which is what it currently does), unlike SoarTech where it's just QOL alignment targets against QOL scenes (and same for VOL). But let me know if that's not correct.
> 
> I don't know that we necessarily do for this Eval, but I was thinking of a way to future proof it and bring their alignment targets out of the code and into config.

Thanks Brian, I'll put together a ticket for this.

nextcen-dgemoets:
> I don't know that we necessarily do for this Eval, but I was thinking of a way to future proof it and bring their alignment targets out of the code and into config.

For what it's worth, their alignment targets weren't hard-coded in our code-- we got them by querying TA1 via their API.  In theory that still works for ADEPT, but I agree we should future-proof ADEPT keep them the same as SoarTech.

nextcen-dgemoets:
One quick thing you can add here is to log to history when the ADM calls `get_available_actions()`.  This was something some of us discussed in Teams last week.  I'm 99% certain that ADMs call this exactly once before every call to take_action/intend_action, but we should still log this API call as well.

brianpippin:
Is this ready to go or are we waiting on additional changes?

nextcen-dgemoets:
> Is this ready to go or are we waiting on additional changes?

Still working on getting it all to work...

nextcen-dgemoets (approved):
I think we finally got it!  Still some refactoring to do, but we can give it to TA2 and they can run eval, training, and piecemeal their scenarios by updating configuration.

nextcen-dgemoets:
One other thing worth mentioning:  when we run SoarTech evaluation scenarios and call `get_session_alignment()`, we get an `alignment_source` with two entries, the first with a scenario_id of the scenario we're running (e.g., `qol-dre-1-eval`), and another with a _training_ version of it, e.g., `qol-dre-1-train`.

I think ST is okay with this, but it looks weird.

PR #105, opened 2024-09-19
ITM-586: Final DRE enhancements to scenario runner/config

nextcen-dgemoets:
Work performed:
- Added ADEPT to the current scenario runner configuration.
- Refactored or replaced indicated areas (TODO) in `start_session`;
- Added `ALWAYS_CONNECT_TO_TA1` (for non-training, non-eval sessions) to configuration; and
- Allowed config vars to have embedded environment variables (Parallax request).

To test:
- First, update your `config.ini` with the new stuff added to the template.
- Run all session types (while connected to TA1 servers):  `eval`, `soartech`, `adept`, and `test`.
- Refer to new logging to see which scenarios/alignment targets are configured to run.
- Update your `config.ini` to remove a subset of ST and ADEPT scenarios/alignment targets, and try again.
- Make sure that your config file is internally consistent (e.g., `ADEPT_EVAL_FILENAMES` is in consistent with `ADEPT_EVAL_MJ_SCENARIOS`).
- Run a training scenarios (`--training`) for both ST and ADEPT.
- Use an environment variable somewhere in config, e.g., `ADEPT_URL=http://localhost:%(ADEPT_PORT)s` 

**NOTE**: Currently, standalone non-test sessions don't work (e.g., `--soartech` and `--adept`, when `ALWAYS_CONNECT_TO_TA1` is False).  This will be addressed in the next ticket, `ITM-563`.

IMO, you don't need to have all of these tests run to completion.  Certainly a full `eval` session should be fully tested.  If you want to short-circuit the server so that it shows you what it's going to run without actually running it, add `raise Exception("Done!")` right before `start_session()` returns (should be at `itm_session.py` line 563).

### DIFF
--- a/swagger_server/config.ini.template
+++ b/swagger_server/config.ini.template
@@ -10,16 +10,18 @@ SCENARIO_DIRECTORY=swagger_server/itm/data/%(EVALUATION_TYPE)s/scenarios/
 ; ADEPT_URL=http://localhost:8081
 ; SOARTECH_URL=http://localhost:8084
 
-; AWS TA1 Default URLs
+; AWS TA1 Default URLs (access Prod AWS servers locally)
 ; ADEPT_URL=https://darpaitm.caci.com/adept
 ; SOARTECH_URL=https://darpaitm.caci.com/soartech
 
-; Prod Default URLs
+; Prod Default URLs (use on production AWS only)
 SOARTECH_URL=http://10.216.38.125:8084
 ADEPT_URL=http://10.216.38.101:8080
 
 SAVE_HISTORY=True
 SAVE_HISTORY_TO_S3=True
+; This applies to `soartech` and `adept` non-training sessions
+ALWAYS_CONNECT_TO_TA1=True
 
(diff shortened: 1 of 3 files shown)


phile-caci (approved):
Everything looks good on my end, ran through all the client commands with config having all scenarios and targets and got what I expected in history
```
python itm_minimal_runner.py --name foobar --session eval  
python itm_minimal_runner.py --name foobar --session adept  
python itm_minimal_runner.py --name foobar --session soartech  
python itm_minimal_runner.py --name foobar --session adept --training  
python itm_minimal_runner.py --name foobar --session soartech --training  
```
Then cleared the history folder and went through a few iterations of 3,2,1 scenarios with 3,2,1 targets, and got all the files I was expecting in history from a training session
```
python itm_minimal_runner.py --name foobar --session adept --training  
python itm_minimal_runner.py --name foobar --session soartech --training  
```

brianpippin (approved):
Got same results as Phi!  LGTM

PR #107, opened 2024-10-08
ITM-625 Add configuration for group targets

phile-caci:
Added a config just for group targets


nextcen-dgemoets (commented):
So this is all good and it works-- but another approach could have been to create a different configuration group within the config file for the group targets.

That is, right now everything uses the `DEFAULT` configuration group.  We could have a `GROUP_TARGET` configuration group with all of the settings you added for this ticket.  Then we would have a single setting (probably in `DEFAULT`) that tells us whether we're using `DEFAULT` or `GROUP_TARGET`.  Or we can have two groupings, `SINGLE_TARGET` and `GROUP_TARGET` which we select via a config value in `DEFAULT`.

Anyway, what you have works and the alternative is probably just more work, but I wanted to mention it in case you didn't want to futz with moving/renaming files around or thought of other advantages to the alternative.  I'm fine with the original approach.

phile-caci:
Hi @nextcen-dgemoets , added command line ability to select groups from the config.ini.  Also add the ability to set the port from command line as well.  I kept the environment variable setup and override it only if the command line argument is given.

Display Help
`python -m swagger_server -h`

Run with Group Target
`python -m swagger_server -c GROUP_TARGET`

Run with Port
`python -m swagger_server -p 8181`

Run with bad Config Group will exit program with message
`python -m swagger_server -c BAD_GROUP`

LMK if there's anything else you'd want me to add to the command line/config setup

(When you're good with everything I'll open the PR to everyone else)

nextcen-dgemoets (changes requested):
Functionally everything looks good and works great.
Just a few mostly editorial suggestions.

README.md
```diff
@@ -71,6 +71,18 @@ pip3 install -r requirements.txt
 python -m swagger_server
 ```
 
+### Optional Usage
+usage: python -m swagger_server [-h] -c CONFIG_GROUP -p PORT
```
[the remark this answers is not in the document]
phile-caci: ahh.. I see, just committed an update.

README.md
```diff
@@ -38,9 +38,9 @@ The following properties can be configured:
 - `SCENARIO_DIRECTORY`
     - default is `swagger_server/itm/data/%(EVALUATION_TYPE)s/scenarios/`
 - `SOARTECH_URL`
-    - default is `http://10.216.38.125:8084`
+    - default is `http://10.216.38.25:8084`
```
nextcen-dgemoets, line 41: It looks like this needs to change from 125 to 25 in the config template.

phile-caci: ahh good catch, done...

phile-caci: My search and replace skills are crap today... done!

nextcen-dgemoets (approved):
Everything has been addressed.  Ready to open up to other reviewers.

kaitlyn-sharo (approved):
Didn't seem to break anything for me

PR #115, opened 2024-12-16
ITM-563: Post-Phase1 enhancements to scenario runner

nextcen-dgemoets:
Changes:
- Added a “test mode” to server command line; remove ALWAYS_CONNECT_TO_TA1 configuration;
- Load alignment target ids at outset but load alignment targets on-demand (based on new sessions);
- Support local runs (no connection to TA1) that only use configured scenarios/targets;
- Support client `--scenario` option (only run configured scenarios);
- Fixed probe alignment bug (ITM-713);
- Removed local alignment targets; get ta1 names from configuration; and
- Added better error messages for the most likely configuration problems (e.g., bad scenario IDs, filenames).

To test, try every session type in both testing and regular mode.  Ensure the evaluation can be run exactly the same as before.
Try making some config errors and see if the error messages are helpful.

Examples:
- regular and testing server mode
- adept session, soartech session, eval session
- solo and full training modes from client
- with config errors
- Running a specific scenario (with `--scenario` in client) to ensure it only runs configured scenarios
- with a subset of alignment targets/scenarios based on configuration

PS:  You should probably set `SAVE_TO_S3` and possibly `SAVE_HISTORY` to false, although checking local history at least once is a good idea to make sure there are no regressions.


kaitlyn-sharo (approved):
Tested everything possible in the client side with test mode and without test mode, seemed to work!

dereknop (approved):

PR #118, opened 2025-01-28
ITM-835: Break model into generic and domain-specific components

nextcen-dgemoets:
Lots of files, but most of the files are generated model files.  Pair with [client PR](https://github.com/NextCenturyCorporation/itm-evaluation-client/pull/66).

To test, run the various session types (`test`, `soartech`, `adept`).  You can start the server in test mode (`-t`) so you don't have to worry about TA1 servers.

Also, take a look at the `Base*` objects and let me know if you think there's anything domain-specific in it.  Note that I think down the line, we'll rename `Threat` to `Risk` or something like that, which is a more generic (not necessarily military) term.  But I'm not doing that now because the current requirement is to remain backward-compatible with P1E scenarios.

In the end, I had to create a [new ticket](https://nextcentury.atlassian.net/browse/ITM-841) for the part of the original ticket where we break up the swagger file into multiple files.  See that ticket for details, if interested.


kaitlyn-sharo (approved):
Makes sense to me, and nothing seemed to break!

dereknop (approved):

PR #119, opened 2025-02-12
ITM-842: Break up server code into generic and domain-specific components

nextcen-dgemoets:
- Break up server code into generic and domain-specific components;
- Add `domain` to `start_session` API
  - Instantiate domain classes at runtime based on specified domain;
- Added cookbook to create a new domain (see "Adding a domain" in `README.md`); and
- Updated `swagger.yaml` to to remove domain-specific content from non-domain-specific areas.

See corresponding [client PR](https://github.com/NextCenturyCorporation/itm-evaluation-client/pull/67).

To test:
- Run an `eval` session type, and a `test`, `soartech`, and `adept` session without connection to TA1 servers.  For these, launch the server with `-t` for test.
   - For the `eval` test, you might want to set `SAVE_HISTORY_TO_S3=False`.
- Perform some tests (maybe with human simulator in client repo) to ensure that returned state looks right, and that scene transitions happen when they should.
- Ensure that code for domain-specific aspects of the TA3 server are encapsulated in separate files / subclasses.
- Peruse all server and client READMEs.


kaitlyn-sharo (approved):
I did not see anything wrong with any of this.

swagger_server/itm/itm_session.py
```diff
@@ -89,8 +93,10 @@ def __init__(self):
         self.itm_scenarios = []
         self.itm_scenario: ITMScenario = None
 
+        # Domain config
+        self.domain_config: ITMDomainConfig = None
         # Action Handler
-        self.action_handler: ITMActionHandler = ITMActionHandler(self)
+        self.action_handler: ITMActionHandler = None
```
dereknop, line 99: What changes here?

swagger_server/itm/itm_scenario.py
```diff
@@ -26,33 +25,20 @@ def __init__(self, yaml_path, session, training = False) -> None:
...
-            else:
-                initially_hidden_injuries = [InjuryStatusEnum.HIDDEN, InjuryStatusEnum.DISCOVERABLE]
-                character.injuries[:] = \
-                    [injury for injury in character.injuries if injury.status not in initially_hidden_injuries]
-                character.unstructured_postassess = None
-                character.vitals = Vitals()
-        for injury in character.injuries:
-            injury.treatments_required = None
-
+    def clear_hidden_data(state: State):
```
dereknop, line 34: I assume this is a placeholder for each domain having it's own version of this function? Is the code that got deleted living somewhere else now that I missed or is it no longer needed at all?

swagger_server/itm/itm_session.py
nextcen-dgemoets: Before, we just statically initialized it to be an `ITMActionHandler`.  Now, this class is subclassed because it has domain-specific behavior, so it it initialized from the `DomainConfig` in `start_session` on line 446:
- `self.action_handler = self.domain_config.get_action_handler(self)`

swagger_server/itm/itm_scenario.py
nextcen-dgemoets: Yes to both questions.  This is the base class (parent class) implementation of clearing hidden data-- which is a null implementation, because none of the basic (non domain-specific) state needs to be hidden at the start of the scene.  Domains may override this method if they have any domain-specific state that needs to be hidden (like triage does).  So the code that got deleted is now in `TriageScenario` (line 17).

dereknop: Ah okay makes sense

dereknop (approved):
Everything made sense and I was able to run each type of session without any issues unless you count eval taking 3 business days.

PR #120, opened 2025-02-18
ITM-855: Break up TA1 specifics into separate config/code

nextcen-dgemoets:
- Moved TA1-specific code and configuration to Adept and Soartech TA1 controllers, children of `ITMTa1Controller`
  - The only place you should see the string `adept` or `soartech` is in these child classes, and in configuration.
- `ITMTa1Controller` is now an abstract base class that contains code shared between TAs and creates TA1-specifc controllers on demand.
  - It is used both for static initialization and session-specific connection to TA1 via TA1-specific instances.
- Added documentation for adding a TA1 to the README, with references to updating the clients.
  - See the corresponding [client PR](https://github.com/NextCenturyCorporation/itm-evaluation-client/pull/68) for this documentation.

To test, run the various session types and make sure everything runs properly.  Run at least one session connected to TA1.  Remember that for tests where you don't need TA1 servers, use the `-t` option on the server.


kaitlyn-sharo (approved):
It appears to still work, changes looked good from my limited knowledge

PR #122, opened 2025-03-13
Support running of ADEPT Phase 1 scenarios with multi targets generated from human participants

nextcen-dgemoets:
***THIS PR IS READY FOR REVIEW BUT WILL NOT BE MERGED***

This branch takes the output from [ITM-892](https://nextcentury.atlassian.net/browse/ITM-892) (which pulls out the KDMAs from human participants to use as targets for ADMs) and creates synthetic alignment targets out of them, such that when ADMs request an `adept` session, the ADEPT Phase 1 scenarios will be served up once per synthetic target.  This is about 500 targets times 3 evaluation scenarios.

The branch expects a file in the server root directory named `text_kdmas.csv`, which is the output from ITM-892 (see [the PR](https://github.com/NextCenturyCorporation/itm-ingest/pull/114)).  For testing purposes, you might want to cull it down to a manageable number of rows.  Make sure your `config.ini` has the same values for `EVAL_NAME` and `EVAL_NUMBER` as the template (`config.ini.template`) and that `SAVE_HISTORY` = True.  You might want to copy the template from this branch, then tweak for your env as necessary.  Then test with the client:
- `python itm_minimal_runner.py --session adept --name itm893`

Note that session alignment is irrelevant here (it will always be NaN).  Make sure that the resulting JSON file contains the TA1 session ID, MJ alignment target, IO alignment target (these are carried forward from the list in ITM-892), the scenario ID, and the MJ and IO KDMA measurements from the TA1 server.

[ITM-892]: https://nextcentury.atlassian.net/browse/ITM-892?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,5 +14,5 @@ EXPOSE 8080
 
 ENTRYPOINT ["python3"]
 
-CMD ["-m", "swagger_server", "-c", "GROUP_TARGET"]
+CMD ["-m", "swagger_server"]
 
--- a/swagger_server/config.ini.template
+++ b/swagger_server/config.ini.template
@@ -26,8 +26,8 @@ ALWAYS_CONNECT_TO_TA1=True
 HISTORY_DIRECTORY=itm_history_output
 HISTORY_S3_BUCKET=itm-ui-assets
 
-EVAL_NAME=Phase 1 Evaluation
-EVAL_NUMBER=5
+EVAL_NAME=Phase II E1
+EVAL_NUMBER=7
 
 SOARTECH_EVAL_FILENAMES=phase1-soartech-eval-qol2.yaml,
     phase1-soartech-eval-qol3.yaml,
--- a/swagger_server/itm/itm_alignment_target_reader.py
+++ b/swagger_server/itm/itm_alignment_target_reader.py
@@ -7,7 +7,9 @@ from swagger_server.models import (
 class ITMAlignmentTargetReader:
     """Class for converting YAML data to ITM scenarios."""
 
-    def __init__(self, yaml_path: str):
+    COUNTER: int = 1
+
+    def init_from_yaml(self, yaml_path: str):
         """
         Initialize the class with YAML data from a file path.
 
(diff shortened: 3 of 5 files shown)


nextcen-dgemoets:
NOTE: At the moment, this code is not clever enough to spot duplicate runs (the same MJ and IO kdma value).  The downside is longer execution time.  LMK if that is desired/requested.  I can hack in some code to save MJ-IO combos it's already done, and skip those scenarios.

kaitlyn-sharo:
> The script worked for me, and the JSON seems to have the session ID, scenario ID, targets, and kdma scores. I know you mentioned expecting `NaN` as the alignment score, but my vscode seems pretty unhappy with that since `NaN` isn't valid JSON. Maybe we could use null instead. It isn't a big deal so if you wanna keep it as is, I have no qualms.

Actually I was thinking this too, but since you had mentioned the NaN in the description, I decided not to say anything.

nextcen-dgemoets:
> > The script worked for me, and the JSON seems to have the session ID, scenario ID, targets, and kdma scores. I know you mentioned expecting `NaN` as the alignment score, but my vscode seems pretty unhappy with that since `NaN` isn't valid JSON. Maybe we could use null instead. It isn't a big deal so if you wanna keep it as is, I have no qualms.
> 
> Actually I was thinking this too, but since you had mentioned the NaN in the description, I decided not to say anything.

Interesting...it's always been this way, but maybe we haven't had much NaN session alignment (which is good).  It's currently just dropping the `AlignmentResults` object into history (which gets dumped to the JSON file), but I guess I can preprocess the `AlignmentResults` and replace a `score` of `NaN` with `None`.  That tweak should find its way into `development` as well.

dereknop (approved):

nextcen-dgemoets:
> Interesting...it's always been this way, but maybe we haven't had much NaN session alignment (which is good). It's currently just dropping the `AlignmentResults` object into history (which gets dumped to the JSON file), but I guess I can preprocess the `AlignmentResults` and replace a `score` of `NaN` with `None`. That tweak should find its way into `development` as well.

Actually, I may not be able to do that, since `score` is required in `AlignmentResults`.  I'll try making a copy of the 
 `AlignmentResults` and dump that to history, and leave the `AlignmentResults` I'm sending across the wire with `NaN`.

nextcen-dgemoets:
Okay, I added the check for duplicates (will run every unique kdma pair exactly three times-- once per scenario) and finally found a way to return null instead of `NaN` for `NaN` scores to keep JSON happy.  I also ran it on the full 498-line output from ITM-892 and got the expected number of processed scenarios.

This should probably be re-reviewed.

dereknop (approved):
LGTM. VS Code is happy with the JSON now

brianpippin (approved):
Nothing jumped out at me, I think I am going to point Kitware at this now and see how things looking with their testing on it.

nextcen-dgemoets:
Closing this PR as it was never intended to be merged.  The work is still available on the [multi-target experiment](https://github.com/NextCenturyCorporation/itm-evaluation-server/tree/multitarget-experiment) branch, and the [one thing](https://github.com/NextCenturyCorporation/itm-evaluation-server/pull/122#issuecomment-2725325213) we wanted to move to `development` has been done [here](https://github.com/NextCenturyCorporation/itm-evaluation-server/pull/129/commits/d40ff31f0e5f7c81c4b991f2568b830fd8d2cf3b).

PR #123, opened 2025-03-17
ITM-841: Migrate to the OpenAPI Generator

nextcen-dgemoets:
Also see the corresponding [ticket](https://nextcentury.atlassian.net/browse/ITM-841) and [client PR](https://github.com/NextCenturyCorporation/itm-evaluation-client/pull/70).

**NOTE**: I suggest testers clone a new server and client repo, then check out this branch and create a new python env, etc.  We're changing gradle versions and dependencies here, so it might be cleaner, and facilitate comparing to the baseline `development` branch.  Remember to create a `config.ini` from the template with your local specifics.

Motivation for this migration was to be able to split our YAML up into multiple files, isolating domain-specific constructs into a separate file.  In short, the approach was to generate code from these YAML files using the OpenAPI generator (v3), replace all model files, fold other code differences into our baseline, and make any other necessary changes to accommodate the new generator.

To test, I would suggest re-generating the models via `./gradlew` and running all session types (with at least one `adept` and `soartech` run integrated with the TA1 server).  Remember for non-integrated tests, you can launch the server with `-t`.  You can also try the human client (`itm_human_input.py`).

Notes:
- When generating the models on Windows you will see `Failed to get the schema name: null`.  According to [their code](https://github.com/OpenAPITools/openapi-generator/blob/31c1a86736cd67b31d884e67779ab63247d26920/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java#L384), this happens when simple ref (e.g., `#/components/schemas/Pet`) is null, but I'm not seeing a case of this.

- The generator can't seem to handle dictionaries where the values are enums (e.g., `character_importance` is a map of string id to `MissionImportanceEnum`)
.  The work-around / compromise was to use a string instead of `MissionImportanceEnum`.

    - We can probably make `character_importance` an object with `id` and `importance` properties, but this is a breaking change to Phase 1 scenarios.  When backward compatibility is no longer required, we can pursue this approach.

- `meta_info` `scene_ids` are causing problems when the id is a simple integer.  The work-around was to change our example that does this, but we've always said that you can name your scenes 0-n and not specify `next_scene` (à là metrics eval).  So this is no longer the case.  Neither TA1 did that in Phase 1, and in fact SoarTech changed their 0-n to "id-0", ... "id-n" at some point, I believe because they were encountering a similar problem.

- client-side `get_swagger_class_enum_values` return metadata (e.g., `"<InjuryLocationEnum.LEFT_CALF: 'left calf'>"`) instead of the string value (e.g., `"left calf"`).  This seems to be just a "display" problem in clients.  If anybody knows how to strip this out when displaying an enum value, let me know.


dereknop:
I had to run `./gradlew` instead of `./gradle`. I did a fresh clone of the server and client. Ran the server connected to TA1 servers and then:
```
python itm_minimal_runner.py --session soartech --name foobar
python itm_minimal_runner.py --session adept --name foobar
```
Then with the server in test mode I ran:

`python itm_minimal_runner.py --session test --name foobar `

I didn't experience any errors or see anything alarming. Looking through the code rn.

README.md
```diff
@@ -158,13 +156,20 @@ Modify the definitions of the domain-specific versions of various state objects
...
 - `DomainActionTypeEnum`: domain-specific (string) action types supported in the server
 - `EntityTypeEnum`: domain-specific enumeration of available entity types; can be a subject or object of a `MESSAGE` action
 
 Add YAML definitions for any nested objects defined in the above state objects.  If any other model objects require domain-specific content, then create both a `Base` and `Domain` version of each object, and change the current object to include both of these versions with the `allOf` keyword. This may also entail other server code changes.  All YAML changes should be made to the `swagger/swagger.yaml` file, and used by your client ADM.  Note that the hope is to move all domain-specific state to a separate file, but this isn't supported cross-platform with our current dependencies.
 
+**NOTE**: Certain enums would ideally be extensible so we can define basic values in `base_swagger.yaml` and domain-specific extensions in `domain_swagger.yaml`.  Until this is supported in the OpenAPI spec and openapi generator, the enums from three domain-specific enums will need to be replicated in the base swagger file:
```
dereknop, line 166: 'OpenAPI spec' vs. 'openapi generator' is throwing me off. I'd ignore it if it wasn't capitalized in the `.openapi-generator-ignore`

swagger_server/controllers/itm_ta2_eval_controller.py
```diff
@@ -112,22 +114,22 @@ def _reclaim_old_session():
...
-    :param session_type: the type of session to start (`eval`, `test`, or a TA1 name)
+    :param session_type: the type of session to start (&#x60;eval&#x60;, &#x60;test&#x60;, or a TA1 name)
     :type session_type: str
     :param adm_profile: a profile of the ADM in terms of its alignment strategy
     :type adm_profile: str
-    :param domain: A domain supported by the ITM evaluation server, as specified in server configuration
+    :param domain: A domain supported by the ITM evaluation server
     :type domain: str
-    :param kdma_training: whether this is a `full`, `solo`, or non-training session with TA2
+    :param kdma_training: whether this is a &#x60;full&#x60;, &#x60;solo&#x60;, or non-training session with TA2
```
dereknop, line 127: Showing as &#x60 for me on github and vscode. I looked it up and see it is unicode for `, wonder how this happened

nextcen-dgemoets: That's exactly how the OpenAPI Generator generates those comments; I just copy-pasted those comment blocks.  If you think it's a basically a bug on their part, LMK and I'll change it to the actual characters.

nextcen-dgemoets:
> I had to run `./gradlew` instead of `./gradle`. I did a fresh clone of the server and client.

That was my mistake in the PR description.  Fixed/changed to `./gradlew`.

swagger_server/controllers/itm_ta2_eval_controller.py
dereknop: No need to change then I suppose, it is insignificant, just caught my eye

dereknop (approved):

README.md
nextcen-dgemoets: Fixed.

kaitlyn-sharo (approved):
Nothing seemed to break

PR #124, opened 2025-03-17
ITM-877:  Consume supplies upon unsuccessful open abdominal wound treatment

nextcen-dgemoets:
Although there’s a larger conversation on how these unsuccessful treatments consume supplies, for now, to match the sim, the ADM API should consume hemostatic gauze and pressure bandages when used by ADMs to treat open abdominal wounds.

To test, run:
- `python itm_minimal_runner.py --session test --scenario supply_consumption --name itm877`

Maybe add `print(state.supplies)` in `itm_minimal_runner.py` line 239 (of `development` branch) so you can check the client output and see that the hemostatic gauze and pressure bandage count went down after their respective attempted treatments.

NOTE:  I also removed the decompression needle treatment from the test, since the behavior it was testing was left on the cutting room floor of its PR but was inadvertently left in the test.

NOTE2:  I've been meaning to close files throughout the server (good practice) so am adding that here.

### DIFF
--- a/swagger_server/itm/data/domains/triage/test/supply_consumption.yaml
+++ b/swagger_server/itm/data/domains/triage/test/supply_consumption.yaml
@@ -342,12 +342,12 @@ scenes:
   - id: fix_kristin
     end_scene_allowed: false
     state:
-      unstructured: Testing applying hemostatic gauze to the untreatable open abdominal injury.
+      unstructured: Testing applying hemostatic gauze and a pressure bandage to the untreatable open abdominal injury.
       characters:
       - id: Kristin
         name: Kristin
         unstructured: >
-          Patient has an untreatable open abdominal injury, but treating with hemostatic gauze will consume the supply.
+          Patient has an untreatable open abdominal injury, but treating with hemostatic gauze or pressure bandanges will consume the supply.
         demographics:
           age: 34
           sex: F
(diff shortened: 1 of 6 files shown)


dereknop (approved):
Worked as you described. To anyone else reviewing `print(state.supplies)` is a lot easier to follow in the console

nextcen-dgemoets:
> Worked as you described. To anyone else reviewing `print(state.supplies)` is a lot easier to follow in the console

Good idea.  Updated PR description.

swagger_server/itm/data/domains/triage/test/supply_consumption.yaml
```diff
@@ -345,3 +338,54 @@ scenes:
...
+          race: White
+        vitals:
+          avpu: ALERT
+          ambulatory: true
+          mental_status: CALM
+          breathing: NORMAL
+          heart_rate: NORMAL
+          spo2: NORMAL
+        injuries:
+          - name: Open Abdominal Wound # Hemostatic gauze will be consumed if applied here
```
kaitlyn-sharo, line 363: This might be a little silly, but here and on line 350, I feel like pressure bandages should be mentioned as well. Maybe "Both gauze types will be consumed" or "treating with either gauze type will consume the supply"?

kaitlyn-sharo (approved):
Seems to work!

swagger_server/itm/data/domains/triage/test/supply_consumption.yaml
nextcen-dgemoets: Not silly!  I had initially forgotten about pressure bandages in this ticket and added it later.  I should have updated these two lines (and line 345) to reflect this.

PR #125, opened 2025-03-19
ITM-715:  Add metadata to server output

nextcen-dgemoets:
Sometimes the first item in "history" is start scenario but other times it is start session, so not all the history files are exactly the same structure.  Duplicating certain fields to the initial evaluation section would make it easier on downstream components:
- scenario name
- scenario id
- alignment target
- adm name
- ta1 name
- ta3 session id
- ta1 session id

To test, try out different session types and check out the `evaluation` section of the JSON output or `metadata` of training `stdout`, some in server test mode (`-t`) and others in regular mode. e.g.:
- `python itm_minimal_runner.py --session test --name test`
- `python itm_minimal_runner.py --session adept --name adept`
- `python itm_minimal_runner.py --session soartech --name soartech`
- `python itm_minimal_runner.py --name adeptrain--session adept --training solo`
- `python itm_minimal_runner.py --name soartrain --session soartech --training full --scenario qol-ph1-train-3`

### DIFF
--- a/swagger_server/itm/itm_history.py
+++ b/swagger_server/itm/itm_history.py
@@ -28,9 +28,23 @@ class ITMHistory:
 
     def clear_history(self):
         self.history.clear()
+        if self.evaluation_info.get('scenario_name'): # Did user set metadata?
+            self.evaluation_info.pop('scenario_name')
+            self.evaluation_info.pop('scenario_id')
+            self.evaluation_info.pop('alignment_target_id')
+            self.evaluation_info.pop('adm_name')
+            self.evaluation_info.pop('ta1_name')
+            self.evaluation_info.pop('ta3_session_id')
+            self.evaluation_info.pop('ta1_session_id')
 
-    def get_history(self):
-        return self.history
+    def set_metadata(self, scenario_name, scenario_id, alignment_target_id, adm_name, ta1_name, ta3_session_id, ta1_session_id):
+        self.evaluation_info['scenario_name'] = scenario_name
+        self.evaluation_info['scenario_id'] = scenario_id
+        self.evaluation_info['alignment_target_id'] = alignment_target_id
+        self.evaluation_info['adm_name'] = adm_name
+        self.evaluation_info['ta1_name'] = ta1_name
+        self.evaluation_info['ta3_session_id'] = ta3_session_id
+        self.evaluation_info['ta1_session_id'] = ta1_session_id
 
     def add_history(self,
                     command: str,
--- a/swagger_server/itm/itm_session.py
+++ b/swagger_server/itm/itm_session.py
@@ -154,7 +154,7 @@ class ITMSession:
         self.state.scenario_complete = True
 
         if self.kdma_training:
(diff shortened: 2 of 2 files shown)


dereknop:
All the data is there as described. I'm not sure that having test for both the `adm_name` and `ta1_name` is the best though. I guess in theory these JSON files probably will never be used for anything since they are test output so it may not matter much. Maybe CACI could be the ta1 name or we could just write `null` since `ta1_session_id` is `null`. 
Screenshot 2025-03-20 at 12 11 58 PM

Example from an adept scenario. Is the `adm_name` adept because I just ran this locally, but for evals it would tell us something like `TAD-Aligned` for instance? 
Screenshot 2025-03-20 at 12 18 47 PM

nextcen-dgemoets:
> All the data is there as described. I'm not sure that having test for both the `adm_name` and `ta1_name` is the best though. I guess in theory these JSON files probably will never be used for anything since they are test output so it may not matter much. Maybe CACI could be the ta1 name since it is our test yaml file or we could just write `null` since `ta1_session_id` is `null`. 
> Example from an adept scenario. Is the `adm_name` adept because I just ran this locally, but for evals it would tell us something like `TAD-Aligned` for instance?

The `ta1_name` comes from the session type (adept, soartech, test, or in the case of eval, either adept or soartech).

The `adm_name` is provided by the ADM, so in "real life" it will be something like `TAD-aligned` or `ALIGN-ADM-OutlinesBaseline`, etc.  For the minimal runner (braindead client), the adm_name is specified on the command line.  I've updated the PR description to avoid this duplication and clarify what these values are / where they come from.  Test sessions don't typically write out to JSON, unless we're testing the JSON output (like in this ticket).

dereknop (approved):
Okay, makes sense 👍

kaitlyn-sharo (approved):
Looks great! This'll make life a lot easier, thanks :)

kaitlyn-sharo:
Could we also potentially move alignment out of history too at some point?

nextcen-dgemoets:
> Could we also potentially move alignment out of history too at some point?

You mean the alignment `score` obtained in the final command of each JSON file?  Sure.
I *did* ask you and Derek if you had any further requests for this section when I was working on the ticket...  \<razz\>

kaitlyn-sharo:
> > Could we also potentially move alignment out of history too at some point?
> 
> You mean the alignment `score` obtained in the final command of each JSON file? Sure. I _did_ ask you and Derek if you had any further requests for this section when I was working on the ticket... <razz>

Well to be fair, I don't really think it belongs in a `metadata` section. I think it maybe should be separated, and then the ta1 session id could be moved over to that section. Also, there's something about seeing it in action that sparked excitement and inspiration :)

nextcen-dgemoets:
> Well to be fair, I don't really think it belongs in a `metadata` section. I think it maybe should be separated, and then the ta1 session id could be moved over to that section. Also, there's something about seeing it in action that sparked excitement and inspiration :)

Added "results" section which contains TA1 session ID (moved from metadata), alignment score, and kdmas.

PR #126, opened 2025-03-25
ITM-497:  Add action validation to API

nextcen-dgemoets:
See [ticket](https://nextcentury.atlassian.net/browse/ITM-497) and [client PR](https://github.com/NextCenturyCorporation/itm-evaluation-client/pull/71).

Exposed server-side action validation code via a simple endpoint.

To test, verify that normal operations still work (`test`, `adept`, `soartech` sessions).  You shouldn't need the TA1 servers, so you can launch the server with the `-t` option.  You also might try hacking the `itm_minimal_client.py` to make some invalid actions (both regular and intent actions) to see that the API catches invalid actions.
For example, near the beginning of the client's `get_net_action()` function, add:
```
        if random_action.intent_action:
            random_action.character_id = 'foobar'
```
or something like that.  Or change `get_random_supply()` to return `foobar` instead.

### DIFF
--- a/swagger/base_swagger.yaml
+++ b/swagger/base_swagger.yaml
@@ -445,7 +445,7 @@ paths:
                 type: string
                 x-content-type: text/plain
           description: An exception occurred on the server; see returned error string.
-      summary: Take an action within a scenario
+      summary: Validate an action within a scenario
       tags:
       - itm-ta2-eval
       x-openapi-router-controller: swagger_server.controllers.itm_ta2_eval_controller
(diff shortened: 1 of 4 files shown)


kaitlyn-sharo (approved):
works as expected

dereknop:
I got all of the usual commands to run with no issues. After sabotaging the client side, I got the following:

```
Action type: ActionTypeEnum.APPLY_TREATMENT; Character ID: foobar; parameters: {'location': 'left stomach', 'treatment': 'Hemostatic gauze'}
Client error: invalid intention generated, message: "Error code 400: ADM character_id (foobar) doesn't match scene (Victim) for action_id 'intend_treat_Victim'.".  Exiting.
```
So I think everything is working as intended.

dereknop (approved):

PR #127, opened 2025-03-28
ITM-488:  Improve simultaneous session handling

nextcen-dgemoets:
See [ticket](https://nextcentury.atlassian.net/browse/ITM-488).

- Support multiple simultaneous sessions with the same adm_name.
   - Use the `session_id` as the “key” and don’t reuse sessions with the same `adm_name` like it used to.
- When the hard cap of active sessions is reached, clear out and re-use a completed session (one that has completed normally)
  - If there are no completed sessions, then clear out and re-use a session that has timed out.
- Externalize session hard cap and session timeout via command-line parameters (with defaults).

To test, launch the server in test mode and set a low hard cap and session timeout, e.g.:
- `python -m swagger_server -t --timeout 1 --max_sessions 1`

Then in the `development` client, add a `sleep(20)` where it changes scenes (e.g., around line 245).

Test the following:
- The server functions normally with the specified number of simultaneous sessions (e.g., `max_sessions`).
- Completed and inactive sessions are reclaimed/recycled as necessary, which the server logs.
- The client gets a 503 "Server Overload" message if a new session is requested but the the max sessions is exceeded and no session can be reclaimed/recycled.
- If a client session times out (i.e., doesn't make a call within `timeout` minutes), it is reclaimed by the server if the max sessions is exceeded.  If the client then makes a call, it should get a 400 "Invalid Session ID" message.
- You can run different sessions with the same `adm_name`.
- The server honors command-line hard caps and session timeouts, but defaults as advertized.
- Anything else you can think of.

**NOTE**: the drawback to this approach (and the reason that a session with the same `adm_name` used to be recycled) is that if you run a bunch of ADM tests create sessions but then fail quickly, you can quickly reach the session maximum and will have to wait for them to start timing out before you can start a session.  We may create an "abort session" API call which a well-behaved client can send if there's a crash/problem (other than a network problem).

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,13 +59,14 @@ The following properties can be configured:
 *NOTE:* the trailing **`s`** in `.../data/%(EVALUATION_TYPE)s/...` is needed for string interpolation to work properly.
 
 ### Scenario filename convention
-Scenario files must be named in the following format to be read by the server at runtime (without punctuation except the indicated hyphens).
-This is the same convention used in the metrics evaluation:
+Scenario files should be named in the following format (without punctuation except the indicated hyphens).
+This was the convention used in the (Phase 1) Dry Run evaluation:
 - `<EVALUATION_TYPE>-<ta1name>-[eval|train]-<id>.yaml`
+At runtime, however, the server will look for the filenames as specified by the `*_FILENAMES` configuration variables.
 
 Please note:
 1. `EVALUATION_TYPE` is the value of the configuration variable defined in `config.ini`;
-2. `ta1name` is either `soartech` or `adept`;
+2. `ta1name` is one of the values from the `ALL_TA1_NAMES` configuration variable;
 3. Use `eval` for evaluation scenarios and `train` for training scenarios; and
 4. The `id` should be derived from the scenario ID in the YAML file, although it isn't required, e.g., `qol`, `MJ2`, `urban`, etc.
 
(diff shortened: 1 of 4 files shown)


dereknop:
> * The server functions normally with the specified number of simultaneous sessions (e.g., `max_sessions`).

I Increased the`max_sessions` to `3` and running 2 simultaneous sessions worked as expected 

> * The client gets a 503 "Server Overload" message if a new session is requested but the the max sessions is exceeded and no session can be reclaimed/recycled.

This worked as you described
`raise ServiceException(http_resp=http_resp, body=body, data=data)
swagger_client.exceptions.ServiceException: (503)
Reason: SERVICE UNAVAILABLE`

> * If a client session times out (i.e., doesn't make a call within `timeout` minutes), it is reclaimed by the server if the max sessions is exceeded.  If the client then makes a call, it should get a 400 "Invalid Session ID" message.

I got the following log when starting my 4th session, when I had a `max_sessions` of 3. I placed a large `time.sleep(200)` in `itm_minimal_runner.py` to make sure I'd have plenty of time to start them up simultaneously:

`INFO 04-02 05:31:34 Reclaiming inactive session 7d254cf6-d469-4446-a166-99a05b020059, name 'foobar'`

And then when the expired session tried to make another call:

`swagger_client.exceptions.BadRequestException: (400)
Reason: BAD REQUEST
HTTP response headers: HTTPHeaderDict({'Server': 'Werkzeug/2.2.3 Python/3.11.4', 'Date': 'Wed, 02 Apr 2025 21:43:16 GMT', 'Content-Type': 'application/json', 'Content-Length': '21', 'Connection': 'close'})
HTTP response body: "Invalid Session ID"`

> * You can run different sessions with the same `adm_name`.

I ran all of my sessions with the same adm name with no issues.

I didn't come across anything that would stop me from approving. I am curious how often an ADM would go a whole hour without making a call to the server and still actually be valid (since the default timeout is an hour).

dereknop (approved):

nextcen-dgemoets:
> I am curious how often an ADM would go a whole hour without making a call to the server and still actually be valid (since the default timeout is an hour).

Probably not under normal circumstances.  David Joy / Kitware suspects that maybe 30 minutes would be plenty long, so I doubled it (or I should say, I didn't change the original default of 1 hour).  If an ADM takes a long time to process results from a previous action, or is waiting for some other component to do something, maybe it will take an hour.  The API in theory is generic, i.e. domain- and ADM-agnostic.  Maybe in some domains (or some ADM implementations) an hour is a typical amount of time for an ADM to respond (let's hope it's not a small unit triage ADM).  That's why I added the `timeout` server command-line parameter so the default of one hour can be overridden.

And keep in mind:  it's not like the server cancels the session after an hour of inaction-- it's just that if the server needs to allocate a new session and its hard cap is met (which is currently 250!), then it could cancel the session so it can be re-allocated to the new session it's trying to allocate.  In practice, this isn't likely to happen.

kaitlyn-sharo (approved):

PR #128, opened 2025-05-05
ITM-864: Create a sample domain with code/test support

nextcen-dgemoets:
Improved multi-domain support including easily switching between domains.  See [ticket](https://nextcentury.atlassian.net/browse/ITM-864) and corresponding [client PR](https://github.com/NextCenturyCorporation/itm-evaluation-client/pull/72).

Changes:
- Added gradle property for rebuilding models for the specified domain;
  - Generate `domain_swagger.yaml` from selected domain's swagger file
- Defined certain enums in domain swagger file until `oneOf` is supported for extensible enums in OpenAPI spec;
- Created new sample domain (Hunt the Wumpus) with placeholder implementation and sample test YAML;
- Added `Unknown` to `RaceEnum` model;
- Updated documentation; and
- Made a couple of minor fixes.

To test:
- The triage domain (tests, evals with or without TA1 servers) should still work out-of-the-box after running `./gradlew` in both repositories.  Test this.
- To switch to the wumpus domain, run `./gradlew -Pdomain=wumpus` in both repos.
  - On Windows, you might want to run `git stash ; git stash pop` so that git doesn't think that every model file has changed because of CR/LF differences.
- You will need to add `,wumpus` to `SUPPORTED_DOMAINS` in your `config.ini`.
  - This is not currently in the template because it will be addressed soon in [ITM-978](https://nextcentury.atlassian.net/browse/ITM-978).  That said, LMK if you think it should be in the template in this PR.
- Run a `test` session with both the minimal and human clients:
  - `python itm_minimal_runner.py --session test --domain wumpus --name wumptest`
  - `python itm_human_input.py --domain wumpus --session test`
  - Start the server with `-t` since the `wumpus` domain does not support a TA1.
- Switch back to the triage (default) domain with `./gradlew`
  - On Windows, you might want to run `git stash ; git stash pop` so that git doesn't think that every model file has changed because of CR/LF differences.
- Run a `test` session with both the minimal and human clients:
  - `python itm_minimal_runner.py --session test --name triagetest`
  - `python itm_human_input.py --session test --scenario reusable_supplies`  (because it's short)
  - Test with a TA1 at least once.
- Feel free to run any other tests!  Help me find my bugs!

Notes:
- As you may be able to tell by `swagger/wumpus_swagger.yaml`, the placeholder `Hunt the Wumpus` domain just adds a `foobar` property in several places in the model hierarchy, and adds a `SHOOT` command.  The `foobar` property at the `Character` level is hidden from the client (like patient vitals).  When testing with the human client in particular, ensure that the properties are there and with/without values as appropriate.
- In the Hunt the Wumpus domain, a `character` is a room in the cave.  See the placeholer `README-wumpus.md` in the client repo.
- In ITM, it is only valid to MOVE_TO a character (or "room") if it is "unseen".  The minimal client prevents moving to a "seen" character, but the human client does not (so that it can be tested).  This means that when running the human client, to complete the scenario, I'd suggest your first `MOVE_TO` action be to Room 3, then in the next scene, your `MOVE_TO` action should be to Room 2.  That said, you can test and see what happens if you don't.

[ITM-978]: https://nextcentury.atlassian.net/browse/ITM-978?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dereknop:
- Ran `./gradlew` in both directories
- Tested triage domain was working as expected with the following from the client while the server was running `python itm_minimal_runner.py --session test --name foobar` and `python itm_minimal_runner.py --session adept --name foobar`. These worked as usual.
- I re ran `./gradlew` with the wumpus domain and added it to the supported domains in config.ini `SUPPORTED_DOMAINS=triage,wumpus` 
>LMK if you think it should be in the template in this PR.

I don't think that is necessary since I'd think the template should be default options, maybe a comment there would be helpful though. 

- With the server running with the `-t` flag, I ran `python itm_minimal_runner.py --session test --domain wumpus --name wumptest` which gave the following output:
Screenshot 2025-05-05 at 3 49 48 PM

- I was able to complete the scenario using the human input tool as well. 
- After running `./gradlew` again, I was no longer able to run the wumpus domain test scenario
- I was able to run an ADEPT session again as well as a test session. 

As far as I can tell, I didn't run into any bugs or issues. I'm not sure I have tested enough to approve yet so I'll play around with the human input tool again and then approve.

dereknop:
BTW you will need to build a GUI for the wumpus domain before I can approve!

dereknop:
Screenshot 2025-05-05 at 4 09 10 PM
This is the behavior I got when trying to move to a room/character I had already done a `MOVE TO` action on in the wumpus domain.

dereknop (approved):
I didn't run into anything that seemed problematic and the code looks good 👍

nextcen-dgemoets:
> This is the behavior I got when trying to move to a room/character I had already done a `MOVE TO` action on in the wumpus domain.

Yep, that looks right.  If they call the validate endpoint first, then they can avoid this and make a different selection.  That would be a cool addition to the human client, actually.

jaudick (approved):

PR #133, opened 2025-06-03
ITM-1011: Config enhancements for June collab

nextcen-dgemoets:
See [ticket](https://nextcentury.atlassian.net/browse/ITM-1011) and [client PR](https://github.com/NextCenturyCorporation/itm-evaluation-client/pull/75).

Work performed:
- Make p2triage the default domain (and clear out old models);
- Add configuration for multi-kdma run;
- Copy `config.ini.template` from domain-specific area in Dockerfile; and
- Parameterize domain, config group, port, and testing switch in Dockerfile (with defaults).

Originally, this ticket was just going to change the default domain to `p2triage`, but while working on it I found the problem with [a previous ticket](https://nextcentury.atlassian.net/browse/ITM-978) that allowed for switching domains, so I subsumed that ticket into this one.

Testing notes:
- Use the [ITM-1011 client branch](https://github.com/NextCenturyCorporation/itm-evaluation-client/tree/ITM-1011).
- To facilitate changing domains, I suggest having two local config files, `swagger_server/config.triage.ini` and `swagger_server/config.p2triage.ini`, then when you're starting a server, make sure `swagger_server/config.ini` is a copy of the appropriate config file.
- In all runs/configs, ensure that `SAVE_HISTORY_TO_S3` is set to `False`.

To test the default (`p2triage`) domain:
- Start in the default (`p2triage`) domain, if you aren't already, via `./gradlew`.
- Launch the server normally via the command line: `python -m swagger_server`.
- Run the minimal client (maybe an `adept` session?) and ensure the server is serving up single-kdma scenarios.
- Launch the server with the `MULTI_KDMA` configuration via the command line: `python -m swagger_server -c MULTI_KDMA`
- Run the minimal client and ensure the server is running the multi-kdma experiment.
  - As of this writing, the production ADEPT server doesn't have the multi-kdma subset scenarios (i.e., everything except `June2025-AF-MF-eval`), so you won't get session alignments for those scenarios-- or you can run the ADEPT server locally and get the scenario files from me if you don't have them.
- Run the server via Docker with default settings:
   - **NOTE**: You will need to (temporarily) change `swagger_server/itm/data/domains/p2triage/config.ini.template` to use the middle set of TA1 URLs ("Prod AWS servers locally") and set `SAVE_HISTORY_TO_S3` to `False`.  You will also have to ensure that a copy of the scenario files is in the default/production location, `swagger_server/itm/data/june2025/scenarios/`.
   - `docker build -t swagger_server .`
   - `docker run -p 8080:8080 swagger_server`
- Run the minimal client and ensure the server is running the single-kdma experiment.
- Run the server via Docker with port and config group overridden:
  - `docker run -p 8081:8081 -e "TA3_PORT=8081" -e "CONFIG_GROUP=MULTI_KDMA" swagger_server`
  - Make sure any local ADEPT server isn't also running on 8081.  Either relaunch it with `--port 8080`, or kill it and run the server with an extra `-e "TEST=-t"`
- Run the minimal client and ensure the server is running the multi-kdma experiment.
   - Your client environment may have to have `TA3_PORT=8081`
- Try anything else you can think of, including seeing how misconfigurations and bad command lines are handled.

To test the old (`triage`) domain:
- Switch to the old (`triage`) domain, via `./gradlew -Pdomain=triage`.
- Copy `swagger_server/itm/data/domains/triage/config.ini.template` to `swagger_server/config.ini` and make any necessary local changes.
- Launch the server in testing mode via the command line: `python -m swagger_server -t`.
- Run the minimal client (with `--session test --domain triage`) and ensure in the JSON output that the `evalName` is `Phase 1 Evaluation`.  This is the evaluation name in the default configuration.  You should also see server logging of the configuration (`DEFAULT`).
- Launch the server in testing mode with the `GROUP_TARGET` configuration via the command line: `python -m swagger_server -c GROUP_TARGET -t`
- Run the minimal client (with `--session test --domain triage`) and check the server logging of the configuration (`GROUP_TARGET`).
- Run the server with the `triage` domain via Docker.  Because the ADEPT server is now configured for Phase 2, run the TA3 server in testing mode so it doesn't make calls to the ADEPT server:
   - `docker build --build-arg domain=triage -t swagger_server .`
   - `docker run -p 8080:8080 -e "TEST=-t" swagger_server`
- Run the minimal client (with `--session adept --domain triage`) and check the server logging for the `DEFAULT` configuration.
- Run the server via Docker with port and config group overridden:
  - `docker run -p 8081:8081 -e "TA3_PORT=8081" -e "CONFIG_GROUP=GROUP_TARGET" -e "TEST=-t" swagger_server`
- Run the minimal client (with `--session test --domain triage`) and check the server logging to ensure it's running the `GROUP_TARGET` configuration.
   - Your client environment may have to have `TA3_PORT=8081`
- Try anything else you can think of, including seeing how misconfigurations and bad command lines are handled.

For good measure as a sanity check, maybe run `./gradlew` and run a `p2triage` eval session to make sure everything's back to normal.

**QUESTION for reviewers:**
Note that I removed the `ENTRYPOINT` in the `Dockerfile`.  I worked on this in early May (for the previous ticket) and I remember looking into this and settling on the solution you see in this ticket.  But I'm not sure this is the only/best way to do this, particularly when you consider the teachings of [this article](https://www.cloudbees.com/blog/understanding-dockers-cmd-and-entrypoint-instructions).  Is it important to keep the `ENTRYPOINT`?  Are there any negative downstream effects of removing it?  And should I be considering docker compose?  If you know Docker well and have any suggestions, I'd be happy to hear them, even if it just results in a new ticket as opposed to changes here.


dereknop (approved):

nextcen-dgemoets:
> >   * **NOTE**: You will need to (temporarily) change `swagger_server/itm/data/domains/p2triage/config.ini.template` to use the middle set of TA1 URLs ("Prod AWS servers locally") and set `SAVE_HISTORY_TO_S3` to `False`.  You will also have to ensure that a copy of the scenario files is in the default/production location, `swagger_server/itm/data/june2025/scenarios/`.
> >   * `docker build -t swagger_server .`
> >   * `docker run -p 8080:8080 swagger_server`
> > * Run the minimal client and ensure the server is running the single-kdma experiment.
> 
> This worked for me. It might be more intuitive if it used the `config.ini` instead of the template, since we use that to fit the needs of what we are doing locally, not the template.

The issue is that end-users are expected to update their `config.ini` to their particular needs, so we don't have that under source control; but we need a clean template in the repo so that when we deploy to prod, it has something under source control to use.  So since prod uses Docker, we have it copy the template.  Or maybe I'm missing the point.

> Had to change line 239 of `itm_minimal_runner` to 8081 instead of 8080, like you mentioned, but it worked as expected.

That shouldn't have been necessary if you set TA3_PORT to 8081 in your client environment.

> As for your Docker question, my initial thought is that using `CMD` and not `ENTRYPOINT` makes sense since we want to be able to have a customizable setup for the server. I am not an expert on the setup here, but if we wanted to be more robust, we could have the user edit `config.init.template` rather than pass flags via the command line. And then we wouldn't need variables in command in the Dockerfile. That would, of course, create a ticket for you to change our setup, so I am not sure if that is worth it.

That's kind of going back to how it was before.  The goal was to parameterize the Dockerfile so that after building the docker, we could run it under different configurations at will without having to edit the config template (or Dockerfile) then push to `main` again.  The template should always be exactly what prod needs, and shouldn't be edited by end-users, only when writing (or in this case testing) PRs.  Or again, maybe I'm missing what you're getting at.

PR #144, opened 2025-07-17
Dev to main for July collaboration release

nextcen-dgemoets:
I have no idea why it's showing so many diffs from `main`.  If you look at, for example, the `config.ini.template` on `main` [here](https://github.com/NextCenturyCorporation/itm-evaluation-server/blob/main/swagger_server/itm/data/domains/p2triage/config.ini.template), you'll see it shows July scenario ids/names.  It's showing all the commits that were squashed into main in the last dev-to-main.  I has something to do with when and how we squash / when we don't / etc.  But I'm not sure why GitHub is seeing so many diffs, and I'm surprised it's not showing conflicts.

In any event, run the eval scenarios, try a few different config groups, and try at least one connected to TA1.
At a minimum, run the `SUBSET_ONLY` and `MULTI_KDMA_SUBSET` server configurations, as those are the ones we'll be running soonest.


nextcen-dgemoets:
I can't seem to resolve the merge conflicts because `development` is protected.  Nuking this PR and `development` and starting fresh.

PR #146, opened 2025-07-23
ITM 1009 and ITM 1065: Eliminate Phase 1 Config Triage Hack and Hardcoding In ADEPT TA 1 Controller

NeilDaniel07:
**Original Tickets:**
- [ITM-1009](https://nextcentury.atlassian.net/browse/ITM-1009?atlOrigin=eyJpIjoiZDUwYTgwY2QzODZmNDViMDg5MjgwZWViNDM0YzY0NDkiLCJwIjoiaiJ9)
- [ITM-1065](https://nextcentury.atlassian.net/browse/ITM-1065?atlOrigin=eyJpIjoiZGU3MWM5NzA5ODZjNDA5MGJlNTA1NGY2MTdhYzNiMWIiLCJwIjoiaiJ9)

**Description:** Previously, the `ADEPT TA1 Controller` file `(swagger_server/itm/ta1/adept_ta1_controller.py)` required that the config file used to start the server maintain a strict structure. The file read and parsed exactly 4 single kdma scenario sets and 1 multi kdma scenario set with corresponding alignment targets. Using a config file that does not have this specific scenario configuration information and names of config variables (ex: `ADEPT_EVAL_K2_SCENARIOS`) resulted in a server side crash when the evaluation client `itm-minimal-runner` was run on the server. Furthermore, the `get_session_alignment_path()` method of this controller file hardcoded checks against the alignment target to choose the corresponding distribution target variable from the config file. Hardcoding in multiple places severely limited the possible structure of config files. This manifested itself in the Phase 1 config file `(swagger_server/itm/data/domains/triage/config.ini.template)`, as in Phase 1, there were only two single kdma scenario types. Thus, the config file had to include useless variables with dummy assignments to prevent the ADEPT TA1 Controller file from causing a server crash. This PR removes all hardcoding mentioned above, introducing a robust dynamic structure. Regex is used to parse scenario, alignment target, and distribution target information in the config file, thus loosening the required structure of config files and preventing the need for any specific number of single kdma scenario or multi kdma scenario information sets to prevent a server crash. This allows us to not only remove the dummy variable assignments in the Phase 1 config file, but ensures that our implementation remains future proof and resilient against different config file formats.  

**Disclaimers:**
- All testing that I performed was done by using the `-t` (testing) mode of the TA3 server. This is because the ADEPT production server was down, and I don't have a local TA1 server configured. Though this should have no effect on `adept` and `test` server sessions, it is important for reviewers to pay special attention in ensuring that the changes made in this branch work without the `-t` flag and with connection to a TA1 server. 
- I am unsure whether changes made for `ITM-1065` work as intended, considering that I was unable to test the functionality of ADM runs where session alignment is generated (due to the above disclaimer). Please attempt to use a different ADM session type (in addition to the ones in the testing directions below) when running the client. This will make sure that changes to this method are properly tested and that session alignment scores are generated as expected. 

**Phase 1 Testing:**
1. Drag the updated `config.ini.template` file from `swagger_server/itm/data/domains/triage/config.ini.template` to the `swagger_server` folder, renaming it `config.ini`. Ensure that this is the active configuration file and that the `SCENARIO_DIRECTORY` variable points to the location of your Phase 1 scenario YAML files.
2. Build the swagger model files required to run Phase 1 scenario files using `gradlew -Pdomain=triage`.
3. Start a local instance of the TA3 server by running the terminal command `python -m swagger_server`. Include the `-t` flag at the end of this command only if the TA1 server is not online at the time you perform testing (or use a local TA1 server if possible). 
4. Navigate to the `itm-evaluation-client` repository. Run the dummy ADM `itm-minimal-runner` using the terminal command `python itm_minimal_runner.py --name itm1009 --session adept --domain triage`. Inspect the terminal output to ensure that the ADM was able to run through all scenario files in the same manner as on the `development` branch (compare which scenarios were run) and ensure that no errors occur on the client nor server side. I recommend piping the terminal output to an outfile to better compare ADM run results. To do this, use the terminal command `python itm_minimal_runner.py --name itm1009 --session adept --domain triage > before.txt 2>&1`. 
5. Repeat step 4, this time setting the --session parameter to `test`. 

**Phase 2 Testing (Regression Testing/Backwards Compatibility Check):**
**NOTE: Many steps are repeated from Phase 1 Testing**

1. Kill the current instance of the server. Replace the Phase 1 config file you used in the above testing protocol by renaming it `config.ini.template`. Next, drag the Phase 2 config file from `swagger_server/itm/data/domains/p2triage/config.ini.template` to `/swagger_server`, renaming it `config.ini`. Just as @nextcen-dgemoets recommended in the `ITM-1053` PR, "you can save time by only enabling 1 or 2 alignment targets per kdma in your config file (commenting out or deleting the rest)".
2. Start a local instance of the TA3 server.
3. Run the terminal command `python itm_minimal_runner.py --name itm1009 --session adept`. Compare terminal outputs to the `development` branch (compare which scenarios were run) and ensure that no errors occur on the client nor server side. Once again, piping the terminal output to an outfile may help in this comparison.
4. Repeat step 3, this time setting the --session parameter to `test`. 
5. Once you have verified that the functionality remains for single kdma scenarios, kill the server once again. Then, restart it using the command `python -m swagger_server -c MULTI_KDMA` to test multi kdma scenarios. 
6. Now, repeat steps 3 and 4 to make sure that the dummy ADM can exercise all multi kdma scenarios in Phase 2 as expected (no changes from the `development` branch).


NeilDaniel07:
> Hm I ran across some errors when trying to run the phase 1 domain again. running phase 2 seemed like it ran just fine. 

I saw the same error when running the `itm_minimal_runner` with the production TA1 server. However, I got the same message when running the client on both this branch and the `development` branch (let me know if this is the same for you), so I don't believe this is likely an error with the work done in this branch. I'm not sure why we're encountering this error though (I initially thought this was because the ADEPT production server was offline). @nextcen-dgemoets, do you have any suggestions?

As for the second set of screenshots (running without a TA1 server), I didn't encounter this error so I'm not sure whether it's present on `development` as well (if you could try it out and let me know, that would be great, since the error doesn't seem to relate to the work done in this branch). My other thought is that I see from your screenshot that the error is with `pydantic`. My package versions are a little different from what's specified in `requirements.txt`, since I was having dependency errors before. If you think it might help, I can create a new virtual environment with your package versions and let you know if I reproduce the error you're seeing.

dereknop:
```
HTTP response headers: HTTPHeaderDict({'Server': 'Werkzeug/2.2.3 Python/3.11.4', 'Date': 'Thu, 24 Jul 2025 21:42:37 GMT', 'Content-Type': 'application/json', 'Content-Length': '80', 'Connection': 'close'})
HTTP response body: "Exception communicating with TA1; is the TA1 server running?  Ending session."
```

I am getting this from my client. Happened with both local adept server and the prod adept server. I will come back to this later tonight and try again with a clean repo. For the phase 1 testing, do I have to rebuild the models in the client repo as well?

nextcen-dgemoets:
@joshuale-caci   To run Phase 1 scenarios with the ADEPT server, you need to use their [Phase 1 server release](https://gitlab.com/itm-ta1-adept-shared/adept_server/-/releases/3.9.9).  I didn't realize anybody was going to try that or I would've mentioned it earlier.

It looks like your second error (standalone) is because you didn't regenerate your models.

@dereknop Yes, you need to regenerate the server and client models whenever you change domain...which is tedious I know.  Coincidentally, David Joy (TA2) and I were discussing on Slack that we'd like to be able to do something like
```
from swagger_client.p2triage.models import (
        State,
        Action,
        Character
    )
```
So that it's easier to switch from one to the other.  Basically, keep both sets of models available at a time.

NeilDaniel07:
> @dereknop Yes, you need to regenerate the server and client models whenever you change domain

Sorry for forgetting to mention that in the testing instructions, that should be fixed now.

NeilDaniel07:
> Order of alignment targets doesn't matter either.

Thank you. I realized the optimization I mentioned doesn't improve time complexity after all and just adds overhead.

nextcen-dgemoets:
> I am unsure whether changes made for ITM-1065 work as intended, considering that I was unable to test the functionality of ADM runs where session alignment is generated (due to the above disclaimer). Please attempt to use a different ADM session type (in addition to the ones in the testing directions below) when running the client. This will make sure that changes to this method are properly tested and that session alignment scores are generated as expected.

As long as the scenario is being run against the correct alignment target(s), then it worked-- it doesn't matter if you're actually making calls to TA1.  For example:
- `python itm_minimal_runner.py --session adept --name adept --scenario July2025-PS1-eval`
Resulted in the following output from the client:
- `Scenario July2025-PS1-eval complete for target ADEPT-June2025-personal_safety-0.5.`
That means the correct alignment target was selected.  I found the same for all scenario types.

nextcen-dgemoets (approved):
LGTM.

I tested `triage` and `p2triage` domains in server test mode, and everything looked good.  The expected scenario-alignment target combinations were served up to the client and fully exercised.  I don't know that an integrated test (with TA1 servers) is really necessary here, but I went ahead and did it for the `triage` domain against both soartech and adept.

I also fixed the p2triage tests and got in a one-line change for @dereknop regarding `evalNumber` being an int instead of a String.

For the second reviewer-- don't bother connecting to TA1.  Also, if you don't want to switch between domains, you can clone another copy of the client and server domains and run the gradle command to switch to the `triage` domain in those directories, so you can leave your original/default one as `p2triage`.

dereknop (approved):
This all is looking good on my end.

nextcen-dgemoets:
> `Failed to get the schema name: null`
> 
> This feels vaguely familiar, but when running `gradlew -Pdomain=triage` from the client, I got this. This isn't an issue, right? @NeilDaniel07 @nextcen-dgemoets. It looks like the models were still generated.

It's normal.  It even happens in an online example I saw when I added this in the first place.

PR #148, opened 2025-09-09
Added September 2025 configuration and tweaks

nextcen-dgemoets:
This contains:
- Changes to the scenario generator and preprocessor for September scenarios
- New config for September for testing and (possibly someday) running ADMs

To test, run a selection of server configuration; you will need the September scenarios in the configuration location.


NeilDaniel07 (commented):
This mostly looked good to me, except for two issues which I've pointed out below. Ignoring these however, it seems as though the core functionality has been achieved. I was able to generate the September 2025 scenario YAML files and exercise the dummy ADM on these scenarios in almost all configuration options without any issues. 

**Concerns:**

- It appears as though some legacy code has accidentally been left in the `csv_preprocessor.py` file. See the lines below. Since subset and redact are no longer valid command line arguments and the corresponding variables being modified in the control logic aren't used elsewhere, I assume it's safe to delete these lines. Doing so resolved the runtime error for me. 
```
if args.subset:
    FULL_EVAL = False
if args.redact:
    REDACT_EVAL = True
```
- When I tried running the ADM with the `OPEN_WORLD` server configuration, nothing was executed. On first glance, I'd assume that this is because the YAML files that are referenced aren't present in `/swagger_server/itm/data/sept2025`. These weren't generated by running the preprocessor on the provided CSV files. I'm assuming that you're aware of this and that we're not worried about `OPEN_WORLD` scenarios for now, but just wanted to point this out in case.

**Suggestion For Future Reviewers:**
To make testing far less tedious, I used the automated testing script I'm working on in ITM-1100. While this branch is still under review, it's capable of running the dummy ADM on all eight configuration options and recording the output in just one terminal command. If you're interested in using this and if it's okay with @nextcen-dgemoets, copy the the `automated_tester.py` and `automated_testing_paths.json` files from that branch into this one. Change the JSON file to accurately reflect your local file paths, then run `python automated_tester.py --group 1 --branch ITM-1102`. In the current state, you will also need to delete `GROUP_TARGET` from the `cfgs` array on line 56 of `automated_tester.py`. You can then inspect the generated ADM run output files created in the root of your repo to provide feedback on this PR.

nextcen-dgemoets:
> * It appears as though some legacy code has accidentally been left in the `csv_preprocessor.py` file. See the lines below. Since subset and redact are no longer valid command line arguments and the corresponding variables being modified in the control logic aren't used elsewhere, I assume it's safe to delete these lines. Doing so resolved the runtime error for me.

Yep, that's a copy-paste error-- thanks for catching.  Removed.

> * When I tried running the ADM with the `OPEN_WORLD` server configuration, nothing was executed. On first glance, I'd assume that this is because the YAML files that are referenced aren't present in `/swagger_server/itm/data/sept2025`. These weren't generated by running the preprocessor on the provided CSV files. I'm assuming that you're aware of this and that we're not worried about `OPEN_WORLD` scenarios for now, but just wanted to point this out in case.

Yep, the open world scenarios are not created from CSVs, and there's only one version of them (originally in June).  So I added `OW` to the `IGNORED_LIST` in the scenario converter.  If ADMs were really going to go through the OW scenarios, they'd probably do that in the context of the June or July server branch.

> **Suggestion For Future Reviewers:** To make testing far less tedious, I used the automated testing script I'm working on in ITM-1100.

Yep-- will look at that as part of my review of ITM-1100, thanks.

PR #182, opened 2026-04-21
ITM-1261: Server and API updates for Feb2026 ADM OW "Part 2"

nextcen-dgemoets:
See corresponding [client PR](https://github.com/NextCenturyCorporation/itm-evaluation-client/pull/83).

Part 1 was to have ADMs run the (almost) original June2025 and Feb2026 scenarios with their quaint binary choice actions (Which patient do you treat) which directly parallel the probes they trained with TA1.

Part 2 changes the scenario asks an omniscient ADM (wrt patients and injuries) to treat and tag all patients, then choose three patients to evacuate, and to align responses based on the 16 MF and AF alignment targets.

For this ticket/PR, in short, I did the following:
- Updated the TA3 API (model classes) to re-introduce certain concepts from ITM Phase 1, most notably the `MOVE_TO_EVAC`, and `TAG_CHARACTER` actions.
- Implemented server-side behavior to support these actions, plus the ability to convert the ADM patient treatment order into probe responses to the original Feb2026 probes, such that it could then obtain session alignment from the TA1 server.
  - Specifically, treating and (re-)tagging updates the patient-level unstructured text; sending a patient to evac makes them unseen so they cannot be (validly) selected again.
- Adapted the Feb2026 OW scenarios for Part 2 by creating two new scenes by hand.  TA2 has agreed to this YAML format / experience, and will be updating their ADMs.
- Updated the itm_minimal_runner and itm_human_scenario_runner scripts in the client repo (see [client PR](https://github.com/NextCenturyCorporation/itm-evaluation-client/pull/83)).

To test, you'll need the custom updated YAML files, `feb2026-desert-openworld2.yaml` and `feb2026-urban-openworld2.yaml`.  I'll share them in Teams or [potentially somewhere else](https://nextcentury.atlassian.net/browse/ITM-1241) by the time you read this.  You shold copy these files to your scenario directory (default `swagger_server/itm/data/feb2026/scenarios`) and your TA1 server (`./openapi/data/scenarios`).

You'll also need to copy the `FEB_OPENWORLD` configuration block from `swagger_server/itm/data/domains/p2triage/config.ini.template` into your own `swagger_server/config.ini`.
Then you should regenerate the model files via `./gradle`.
On Windows you might then need to `git stash ; git stash pop` to get rid of CR/LR differences.

Then finally you're ready to test:
- Start the server in test mode with the Feb OW config:  `python -m swagger_server -c FEB_OPENWORLD -t`
- Run the minimal client, e.g., `python itm_minimal_client.py --session adept --name itm1261`
- Ensure that both scenarios are exercised under all 1D MF and AF alignment targets.
- Make sure your local TA1 server is running, then restart the server without `-t`
  - You first may wish to speed up your day by commenting out most of the alignment targets
- Run the minimal client again, maybe with a name of `itm1261_ta1`
- Ensure both scenarios are exercised, and that this time you get alignment value calculations
  - **NOTE:** If your braindead client randomly doesn't treat any patients, you will get an alignment of `None`.  This is correct and normal, and will be addressed in the next step.
  - Keep running until you get an alignment value so you can verify these results in the JSON output.
- Run the human client, `python itm_human_input.py --session adept`
  - It's a little bit funky, but you get used to it.  You need to start a session (`s`), then start a scenario (`c`), then get available actions (`v`), then take an action (`t`) and fill in any parameters as prompted.  After each action, you need to get available actions again before taking another action.
  - Using the human client, treat and tag every patient (or at least some of them), then take the "end_scene_action" action.  Then take the "evac_patient" action for three different patients.  As you use the client, ensure the following:
    - After you treat a patient, that patient's unstructured text is updated to reflect that he/she is treated.
    - After you tag a patient, that patient's unstructured text is updated to reflect that he/she is tagged.
    - If you re-tag a patient, that patient's unstructured text is updated to reflect the new tag.
    - If you re-treat a patient, that patient's unstructured text is unchanged.
    - You can tag-then-treat, or treat-then-tag, and the unstructured text is updated correctly.
    - Once you've ended the first scene, the TA3 server logs that it's sent probe responses, and you get an alignment value.
    - Once you evacuate a patient with `evac_patient`, that patient's unstructured text is updated to reflect having been moved to evac.
    - You cannot evacuate the same patient twice (you will get a non-fatal `BAD_REQUEST` error).
    - After selecting three patients for evacuation, the scenario will automatically end.
- As always, peruse the code and look for errors, issues, bad comments, bad Python, etc.


dereknop:
> Then you should regenerate the model files via `./gradle`.

I had to run `./gradlew`

> * Run the minimal client, e.g., `python itm_minimal_client.py --session adept --name itm1261`

Think you meant `itm_minimal_runner.py` 


I ran this in ~training~ testing mode, and it appears to work as described. Waiting for my ADEPT server Docker container to rebuild, and then I will continue testing.

nextcen-dgemoets:
> I had to run `./gradlew`

That was buried in the instructions.

> Think you meant `itm_minimal_runner.py` 

Thanks-- fixed.  I make that mistake a lot.  I should just rename the file, hehe.

> I ran this in training mode, and it appears to work as described. Waiting for my ADEPT server Docker container to rebuild, and then I will continue testing.

To clarify, the server `-t` is "testing" mode, not to be confused with running a "training" session, which is specified by passing "full" or "solo" to the `kdma_training` parameter to the `/ta2/startSession` endpoint, which in our clients is done via `--training [full | solo]`.  \<Pushes glasses back up on nose.\>

dereknop (approved):
Image
Looks good after running connect to TA1 server. Human input worked as you described.

swagger_server/itm/domains/p2triage/p2triage_action_handler.py
```diff
@@ -35,18 +39,83 @@ def validate_domain_action(self, action: Action, character: Character):
...
+        if action.action_type == ActionTypeEnum.TAG_CHARACTER:
+            # Requires category parameter
+            if not action.parameters or not 'category' in action.parameters:
+                return False, f'Malformed {action.action_type} Action: Missing `category` parameter', 400
+            else:
+                allowed_values = get_swagger_class_enum_values(CharacterTagEnum)
+                tag = action.parameters.get('category')
+                if not tag in allowed_values:
+                    return False, f'Malformed {action.action_type} Action: Invalid Tag `{tag}`', 400
+        elif action.action_type in [ActionTypeEnum.TREAT_PATIENT, ActionTypeEnum.MOVE_TO_EVAC]:
```
dereknop, line 58: Feel free to ignore since it works. Instead of having a`elif` branch that serves to block the `else` branch, it may be better to do something like

```
elif action.action_type not in [ActionTypeEnum.TREAT_PATIENT, ActionTypeEnum.MOVE_TO_EVAC]:
    return False, f'Invalid action_type `{action.action_type}`', 400

```

And then remove the else branch

nextcen-dgemoets: I see what you're saying-- it _is_ less code.  I kind of like the idea of having a little section of code for each action type in the if-then-else-- even if there's nothing to do for some cases-- then ending with a final else that returns an error.  It's more noticeable in the file it was adapted from, `triage_action_handler.py`, which does the equivalent thing for the phase 1 triage domain.  It's more obvious there (because there are more action types) that we're covering all of the action types explicitly, even if there's nothing to do.  The base class `itm_action_handler.py` does the same thing, although there aren't many action types to handle there either.

Garycheng92 (approved):
Looks good, it took a bit longer to test because I forgot to comment out most of the alignment targets but the human input appears to be working, the tagging and treating are as described in the PR and the `evac_patient` text is showing to have been in evac.

PR #185, opened 2026-05-04
ITM-1301: Improve server logging to facilitate debugging multiple simultaneous connections

nextcen-dgemoets:
For this, I added a `log_id` (the last/right-most hex block from the `session_id`) to most server logging.

To test:
- Clone a temp copy of the client repository, and copy your `config.ini` to this repo.
- Start your TA1 server.
- Start the server with an appropriate server config (e.g., `-c FEB_OPENWORLD`).
- Start both clients simultaneously (or staggered slightly) and ensure that you see the log_id prepended to most server logging.
- Maybe peruse server code and see if there are any instances you think adding this additional logging will help with debugging multiple simultaneous connections.

### DIFF
--- a/swagger_server/itm/domains/p2triage/scenario_converter.py
+++ b/swagger_server/itm/domains/p2triage/scenario_converter.py
@@ -342,7 +342,7 @@ if __name__ == '__main__':
         VERBOSE = True
     if args.evalname:
         EVALUATION_NAME = args.evalname
-        OUT_PATH.replace(DEFAULT_EVALUATION_NAME, EVALUATION_NAME)
+        OUT_PATH = OUT_PATH.replace(DEFAULT_EVALUATION_NAME, EVALUATION_NAME)
         for kdma_info in kdmas_info:
             kdma_info['filename'] = kdma_info['filename'].replace(DEFAULT_EVALUATION_NAME, EVALUATION_NAME)
     if args.no_output:
--- a/swagger_server/itm/itm_scenario.py
+++ b/swagger_server/itm/itm_scenario.py
@@ -110,7 +110,7 @@ class ITMScenario:
             try:
                 self.ta1_controller.post_probe(probe_response=response)
             except exceptions.HTTPError:
-                logging.exception("HTTPError from TA1 posting probe.")
+                logging.exception("%s: HTTPError from TA1 posting probe.", self.session.log_id)
             try:
                 # Get and log probe response alignment if not training and TA1 supports it.
                 if not self.session.kdma_training and self.ta1_controller.supports_probe_alignment():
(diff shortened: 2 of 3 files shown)


NeilDaniel07 (commented):
When testing this with two simultaneous client runs, the core behavior and printing to `stdout` looks race-free and correct. Interleaved session logs were much easier to attribute to individual client sessions by the `log_id`. 

> * Maybe peruse server code and see if there are any instances you think adding this additional logging will help with debugging multiple simultaneous connections.

Here are a couple of places where I believe consistent logging would help further. 

1. Some per-session logged exceptions do not include `log_id`, such as the TA1 startup exception logs in `itm_session.py` and related probe/alignment exception handling. Adding the `log_id` prefix here could make concurrent client session failures easier to trace.

2. A few logs still use only the full `session_id`, such as session mapping, reclaim, session end, and history-save/S3 logs. These became ambiguous to me when running multiple clients that were exercising scenarios and writing output at the same time.

I believe prepending `log_id` in these places could significantly assist developers when debugging or tracing through failed client sessions or issues saving ADM decision results. Please let me know your thoughts on this. 

For your reference, these were the exact search commands I used to find the remaining candidates for adding `log_id`:

`git grep -nE 'logging\.exception\("(HTTPError from TA1|Exception communicating|Exception getting|Exception loading)' -- swagger_server/itm/itm_session.py swagger_server/itm/itm_scenario.py`

`git grep -nE 'Saving session_id|Reclaiming|Session .* ended|Saving history|Saving history to S3|Unable to save history|Could not save JSON' -- swagger_server/controllers/itm_ta2_eval_controller.py swagger_server/itm/itm_session.py swagger_server/itm/itm_history.py`

nextcen-dgemoets:
> 1. Some per-session logged exceptions do not include `log_id`, such as the TA1 startup exception logs in `itm_session.py` and related probe/alignment exception handling. Adding the `log_id` prefix here could make concurrent client session failures easier to trace.

It seems like if these happen to one session it will happen to all of them, but in those cases where there's a blip or a bubble, it would be useful to know which ADM was affected.  I added the `log_id` to a number of error conditions.  I can't add it to the static methods because the `log_id` isn't available.

> 2. A few logs still use only the full `session_id`, such as session mapping, reclaim, session end, and history-save/S3 logs. These became ambiguous to me when running multiple clients that were exercising scenarios and writing output at the same time.

The session mapping, reclaim, and session end log the ADM name, which provides identifying information.  Plus I don't think we have access to the `log_id` here, do we?

As for the history/S3 logging, I believe the ADM name is in the `full_filepath` and any other errors should happen directly after these are logged.  What did you find ambiguous?  Maybe I'm not seeing it.

NeilDaniel07 (approved):
> It seems like if these happen to one session it will happen to all of them, but in those cases where there's a blip or a bubble, it would be useful to know which ADM was affected. I added the `log_id` to a number of error conditions. I can't add it to the static methods because the `log_id` isn't available.

That makes sense. I pulled the latest commit and it looks like the exception paths I was concerned about now include `log_id`, including the TA1 startup and probe/alignment cases. I also saw the same behavior in an actual TA1 server startup failure I created. Each failed session had a distinct `log_id`, which added significant clarity to the sequence of failures that occurred. I believe this will add real debugging value.

> The session mapping, reclaim, and session end log the ADM name, which provides identifying information. Plus I don't think we have access to the `log_id` here, do we?

That makes sense, thanks for the explanation. Taking another look, I agree that the mapping and reclaim logs have enough identifying information since they include the full ADM name. It looks to me as though session end logs the full `session_id` rather than ADM name, but this still gives us enough information since the `log_id` can be derived from the `session_id`. I no longer see a case in the relevant output logs where a line cannot be attributed to a particular ADM/session.

> As for the history/S3 logging, I believe the ADM name is in the `full_filepath` and any other errors should happen directly after these are logged. What did you find ambiguous? Maybe I'm not seeing it.

You're right, apologies for missing this detail. The ADM name is present in the history file path. I re-ran a test with two concurrent client sessions and the history lines were in fact distinguishable using the filenames.

Everything looks good to me now. Thank you for your clarifications.

dereknop (approved):

PR #188, opened 2026-05-22
ITM-1358: Combined binary/trinary server configurations

nextcen-dgemoets:
TA2 has asked for server configurations that combine binary/trinary probes/scenarios.  Arguably, I should have in the first place.  I had in my mind (either right or wrong) that they needed to be separate for one reason or another.

You'll need to incorporate the changes to the config template into your local config.ini.  You can probably more-or-less copy the entire contents of `swagger_server/itm/data/domains/p2triage/config.ini.template` to `config.ini`, but make sure your `ADEPT_URL` is set to your local server and `SCENARIO_DIRECTORY` points to your scenario directory (if not the default location).  Also, set `SAVE_HISTORY_TO_S3=False`.  For the tests with TA1, you might want to set e.g. `ADEPT_K1_ALIGNMENT_TARGETS=Jun2026-MF-5` instead of the whole list; same with the other alignment targets.

Make sure your local TA1 server is running, or you can point to production.

Now you're ready to test the scenarios:
- The automated tester provides some high-level status, but you can follow along with `tail -f <filename>` if you like.
- Run `python automated_tester.py --group test --branch 1358`
  - Check client and server output in `./automated_test_results/1358/*test*.txt`.
- Run `python automated_tester.py --group train-solo --branch 1358`
  - Check client and server output in `./automated_test_results/1358/*train-solo*.txt`.
- Run `python automated_tester.py --group ta1-observe --branch 1358`
  - Check client and server output in `./automated_test_results/1358/*ta1-observe*.txt`.
- Run `python automated_tester.py --group ta1-eval --branch 1358`
  - Check client and server output in `./automated_test_results/1358/*ta1-eval*.txt`.
- Run `python automated_tester.py --group train-full --branch 1358`
  - Check client and server output in `./automated_test_results/1358/*train-full*.txt`.


nextcen-dgemoets:
It turns out they only want combined configs for training, which isn't going to happen in time anyway.

